PR #1614, opened 2013-04-08
File Action : Move and Sym Link

nervling:
Hi there,

I tried to post this as a feature request on the forum, but unfortunately it kept complaining about some thread prefix and wouldn't let me post, so I'm just sending this here.

The problem is that some NAS do not support hard linking, in particular unRaid. As a result, the only option to rename and continue seeding are symbolic links. However, I think most people would prefer to have the symbolic link in the original directory, and the file itself get moved to their Movies directory. Meaning Seeding -> Movies, instead of the current Movies -> Seeding.

I added a few lines to create a new File Action, to move the file, and then create a symbolic link in its original location. This seems safer, as it allows people to remove the links when done seeding without breaking their moving collection.

Perhaps there is a better name for this option, or a work around I didn't think of.

### DIFF
--- a/couchpotato/core/plugins/renamer/__init__.py
+++ b/couchpotato/core/plugins/renamer/__init__.py
@@ -117,7 +117,7 @@ config = [{
                     'label': 'File Action',
                     'default': 'move',
                     'type': 'dropdown',
-                    'values': [('Move', 'move'), ('Copy', 'copy'), ('Hard link', 'hardlink'), ('Sym link', 'symlink')],
+                    'values': [('Move', 'move'), ('Copy', 'copy'), ('Hard link', 'hardlink'), ('Sym link', 'symlink'), ('Move & Sym link', 'move_symlink')],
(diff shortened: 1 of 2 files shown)


thephilipjones:
Sorry to reawaken an old thread, but I was searching for exactly this functionality, and I'm not understanding if it still exists in CP. Perhaps there are cascading actions going over my head, but the current version offers just "Move" "Copy" and "Link" options for both Default File Action and Torrent File Action. I'm running on a Synology 415+ at commit abfd9d08 if it matters.

How can I "Move & Symlink" to allow deletion of the downloaded folders when they're no longer needed to seed? Thanks for all the work!

scdemanett:
@thephilipjones Setting the Renamer file action to Link will do this, that is how I have it set on a Drobo 5N. It moves the file and tries to make a hard link for seeding failing that it tries a symlink if it fails all that it reverts to copy.

thephilipjones:
@scdemanett Great, thanks! I was hoping it was a "simplified UX" thing that it's managing for me. I saw that note, but didn't read it to include moving the files beforehand. Just making sure I can clean up my Downloads directory without wreaking havoc on the nicely organized Movies directory.

PR #1672, opened 2013-04-28
Seeding support

mano3m:
Full support for seeding torrents and processing around it.

Note: I had limited time for testing. I did test it with a few torrents for both uTorrent and Transmission, but as this is a significant change we might want a few people to test it?

Design intent:
- Option to turn seeding support on or off
- After torrent downloading is complete the seeding phase starts, seeding parameters can be set per torrent provide (0 disables them)
- When the seeding phase starts the checkSnatched function renames all files if (sym)linking/copying is used. The movie is set to done (!), the release to seeding status.
- Note that Direct symlink functionality is removed as the original file needs to end up in the movies store and not the downloader store (if the downloader cleans up his files, the original is deleted and the symlinks are useless)
- checkSnatched waits until downloader sets the download to completed (met the seeding parameters)
- When completed, checkSnatched intiates the renamer if move is used, or if linking is used asks the downloader to remove the torrent and clean-up it's files and sets the release to downloaded
- Updated some of the .ignore file behavior to allow the downloader to remove its files

Known items/issues:
- only implemented for uTorrent and Transmission
- text in downloader settings is too long and messes up the layout...

To do:
- implement for other torrent downloaders
- complete download removal for NZBs (remove from history in sabNZBd)
- failed download management for torrents (no seeders, takes too long, etc.)

Updates:
- Added transmission support
- Simplified uTorrent
- Added checkSnatched to renamer to make sure the poller is always first
- Updated default values and removed advanced option tag for providers
- Updated the tagger to allow removing of ignore tags and tagging when the group is not known
- Added tagging of downloading torrents

### DIFF
--- a/couchpotato/core/downloaders/sabnzbd/__init__.py
+++ b/couchpotato/core/downloaders/sabnzbd/__init__.py
@@ -43,10 +43,10 @@ config = [{
                 },
                 {
                     'name': 'remove_complete',
-                    'label': 'Remove torrent',
+                    'label': 'Remove NZB',
                     'default': True,
                     'type': 'bool',
-                    'description': 'Remove the NZB from SabNZBd history after it completed.',
+                    'description': 'Remove the NZB from history after it completed.',
                 },  
                 {
                     'name': 'delete_failed',


elneruda:
Hi, I don't know if i'm in the right section to ask, is it possible to add "Symbolic link reversed" as Sickrage Post-processing option ?

PR #3814, opened 2014-08-17
Added putioDownloader

dumaresq:
I have added one file couchpotato/core/downloaders

This file is a downloader for put.io.  I have tested and it does work, though it could use some work.  I read the docs on how to add stuff to couchpotato, but it seems to be related to people who want to fix bugs, in this case I'm not fixing a bug I'm adding a new feature...

Let me know if you need more information.

It does seem that this is a feature requested in #3079

### DIFF
--- a/couchpotato/core/downloaders/putioDownload.py
+++ b/couchpotato/core/downloaders/putioDownload.py
@@ -2,6 +2,7 @@ from __future__ import with_statement
 import os
 import traceback
 import putio
+import shutil
 
 from couchpotato.api import addApiView
 from couchpotato.core.event import addEvent


oimctony:
Hello, i may have spoken too fast. 
I can indeed snatch a movie with CP and see it on my put.io account, but CP doesnt download it on my PC. What i am missing here ?

dumaresq:
There is a check box in the settings "download" this will download file,
but there are a few notes:

1) You need to have couch listen on an external IP that putio can call
 (that's what the callback host is for).
2) It's a little unreliable the socket that gets opened doesn't always get
the whole file, so make sure your download is complete before deleting it
from put.io.

I am trying to come up with better ways to handle this, but so far this is
the only way available.   You are actually better off using firefox with
"downloadthemall" to download as this is reliable, and faster.

oimctony:
Ok ! Normally the external IP works just fine as i can connect to it from my phone in 3G. 
The format of the callback host is an IP address like: 90.x.xxx.xx:1066 . No need for http: etc, right ? 

Is there any way to get some logs for the download ?I can't seem to find anything in the global logs and I want to understand where the problem comes from, as i think the socket doesn't open correctly on my PC. 

Actually, on Windows if you want more reliable "downloader" for put.io, there is this : http://sourceforge.net/projects/putiodwnldmngr/ .
It is a downloader that automatically looks at folder on your put.io account, and download everything there is in it. 

Maybe a way to do so would be to put the CP uploads in a folder dedicated to CP on put.io, and then this software would download it automatically from put.io. 

Thank you for your amazing work btw, i'm in love with CP since years, and i was disapointed not to be able to use it anymore when i moved to put.io :).  Hope i can make it work !

dumaresq:
You might have found a bug. I never tried it except on port 80.  There are
logs, but the download is started on the putio side. So if it doesn't work
with a port number you would never get the log.

As for the downloader whatever I use needs to work on all oses. But if you
have that already setup. Just turn on couchpotatos renamed and point it to
the download location your downloader.  This should just work.

I will check out the callback host with a port number.

oimctony:
Thanks for your answer ! :) 

I'll leave it like this for now as it works quite well, the default download folder for put.io is still a small issue, but more than manageable. 

Small question, do you know where in the code i could specify a dedicated folder to the Couchpotato uploads ?  Could be easier this way. 

Thanks !

dumaresq:
Sorry I don't follow your question.

oimctony:
Sorry, wrongly asked.

Right now, CP puts the torrent file in the "home" folder on put.io. i  want to be able to change that and specify a dedicated folder in put.io for CP uploads, so i was asking if i could specify a folder by myself directly :).

Thanks !

dumaresq:
Ah ok.  Right now it is line 44 in main.py in core/downloaders/putio  I
believe you would pass in parent_id='directory'  I haven't tried yet, and
I'm not 100% what all the params are to the putio API yet.  If it works let
me know and I'll add it to the code...

dumaresq:
Ok I found a bug, in the callback url that does impact when you use a port
number.  I'll patch this.  I also noticed that the parent_id thing isn't
going to work because putio has changed there API.  I'll update the API on
the couch side and add that as a feature.

I'll try to do the change today, but it might take until next weekend for
me to sort out the API changes that are needed.

oimctony:
Thanks for the news !  
I'll wait for the update then :) 

Have a good day !

dumaresq:
Ok I added a the folder feature and fixed the bug in the callback URL.
I've created a pull request (#4456) for RuudBurger, when it's merged you
should be good to go...

oimctony:
You're a rockstar ;) 

Enjoy your day !

bslayton:
Does anyone know if there is a simple way to get this downloader to download the mp4 converted version from put.io, if the original video is not mp4? Or is this a feature that is missing from the downloader?

dumaresq:
It wouldn't be able to do that, because the file name would be different.
If you want to do that you should tell couchpotato not to download it.
Then convert it and download it manually.

I'll think about how to manage this, but I'm not even aware of how I could
remotely kick off the conversion.

leolobato:
There is an API to trigger the MP4 conversion, maybe that could work?

https://put.io/v2/docs/files.html#convert-to-mp4

dumaresq:
Ok that would help convert, but it doesn't return the name or ID of the
file once it is done, so I don't know how useful that would be.

bslayton:
I'm not sure why it isn't clearly documented in the API, but in order to download the converted video it is nearly the exact same as downloading the original file.

Instead of
`/files/<id>/download`
you would do
`/files/<id>/mp4/download`

Adding to what @leolobato said, In addition to using the api to request a conversion, you can also check the conversion status. 

When I get a chance, I will take a look at your code. I don't know python very well. Last thing I want to do is go in a screw it up. :)

PR #4199, opened 2014-11-09
Added button to remove files when deleting media

wolph:


RuudBurger:
Sorry for my late response.
I feel this needs a lot of testing and checking for security issues and faulty loops etc. I don't want this to be responsible for removing stuff that aren't movies ;)

WoLpH:
We can always add a huge disclaimer to the delete button ;)

Better solution might be to list the files that are going to be deleted but
that's a bit out of my scope.

xenithorb:
What happened to this? I started deleting things thinking that it would remove them from the filesystem, and then figured maybe I had filesystem permissions wrong, spent a bunch of time tailing logs and trying to troubleshoot what I assumed was the functionality of the word "delete" (I mean really, come on)  

Where is this feature? Seems like you've been going back and forth on this for 2 years or so...

blade316:
+1

piotrasd:
+1

BDeveau:
+1

GamerGun:
+1

PR #4456, opened 2015-01-04
Bug fixes and new features for putio

dumaresq:
I fixed a bug with the callback url and added a folder option.  Both tested and seem to work correctly.

Thanks to oimctony  for finding the bug.

### DIFF
--- a/couchpotato/core/downloaders/putio/__init__.py
+++ b/couchpotato/core/downloaders/putio/__init__.py
@@ -29,6 +29,11 @@ config = [{
                     'advanced': True,
                 },
                 {
+                    'name': 'folder',
+                    'description': ('The folder on putio where you want the upload to go','Must be a folder in the root directory'),
+                    'default': 0,
+                },
+                {
                     'name': 'callback_host',
                     'description': 'External reachable url to CP so put.io can do it\'s thing',
                 },
--- a/couchpotato/core/downloaders/putio/main.py
+++ b/couchpotato/core/downloaders/putio/main.py
@@ -28,20 +28,32 @@ class PutIO(DownloaderBase):
 
         return super(PutIO, self).__init__()
 
+    def convertFolder(self, client, folder):
+        if folder == 0:
+            return 0
+        else:
+            files = client.File.list()
+            for f in files:
+                if f.name == folder and f.content_type == "application/x-directory":
+                    return f.id
+            #If we get through the whole list and don't get a match we will use the root 
+            return 0
+
     def download(self, data = None, media = None, filedata = None):
         if not media: media = {}
         if not data: data = {}
 
         log.info('Sending "%s" to put.io', data.get('name'))
         url = data.get('url')
-
         client = pio.Client(self.conf('oauth_token'))
+        putioFolder = self.convertFolder(client, self.conf('folder'))
+        log.debug('putioFolder ID is %s', putioFolder)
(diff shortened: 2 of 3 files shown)


bslayton:
I'm a little confused as to why there is a need for a callback host for putio to ping? I'm looking at the downloader code. Can someone point me to some docs in regards to downloader development?

Thanks!

bslayton:
Actually, I see what's going on with the callback url. Instead of polling to see if a transfer has been completed, you're passing in the callback url to notify CP to go and download the file.

dumaresq:
The call back is how putio tells couch that the download has finished.

superguigui:
Hi and thank you for this very nice feature :)
I have a little trouble understanding how the callback works, do we need to open our Couchpotato instance to the web (by ip for an example) ? So the callback value should look something like http://IP:PORT ?
What if we have a password protected Couchpotato instance ? 
For now the snatching part seems to work (the movie is uploaded in put.io) but nothing happens after this, i think I may have used the wrong callback URL but I'm not sure what it should be ^^; (sorry about this).

Thanks again for your work !

dumaresq:
You are correct. Currently the only way putio can tell couchpotato the
download is complete is via the callback. You would need to expose it to
the internet. Don't worry about the password. Assuming you are using the
couchpotato password (not an extra webserver password) it will use the api
key so the password will not be needed.

In truth you probably don't want to use the feature at all. The downloads
are slow and often die without completing (due to connectivity issues to
putio). You are much better off using something like downloadthemall with
Firefox because it does multiple streams and auto retries.

At some point I would like to introduce the ability to call a download
manager but that means doing some cross platform work. I also want to put
some error checking and retry logic but so far this is over my head.

superguigui:
I see, thanks a lot for your answer !

xtreme22886:
Hi @dumaresq 

I know this is somewhat of an older topic but playing around with Put.io and am having the same issue of my file not downloading from Put.io...or maybe it is but very slowly? Anyways, I've been doing some research and I came across an example Put.io client on their API documentation (https://api.put.io/v2/docs/gettingstarted.html#api-clients (very bottom under API Clients)). If you click on the link there it'll link you to their Python script (https://github.com/cenkalti/putio.py/blob/master/putio.py). After skimming over the script, it looks like when they download a file from, they check to see what the file size is and breaks it up into chunks and transmits those chucks. Apparently this speeds up file downloads of larger files.

I wish I could incorporate their download code into your code but I'm simply not that good. Is this something you wouldn't mind visiting and trying to do? I know you mentioned you might want to fix this portion eventually.

Thanks for your time!

dumaresq:
I will take a look when I have some free time.  I don't believe this
feature was available when I created the plugin.  Either way it's worth
mentioning that I don't recommend having couchpotato do the downloads it's
unreliable, and doesn't really know when the connection is terminated or
finished.  I personally let couchpotato tell put.io to start the download
and then I use firefox with downloadthem all to get the files from putio,
this is faster because it is supports multiple streams and more reliable
because it can restart and knows when failures happen.

dumaresq:
There was a bug on the putio server side that I reported to them and they
fixed, downloads should be working now.  I've also updated the couchcode to
use there new method, but I'm got to would with the couchpotato maintainer
in order to incorporate it because of some complications.  I'll update you
when the code is merged back in.

dumaresq:
The updated code seems to have been merged...

PR #4462, opened 2015-01-05
New algogithm to use a crowd logic on newznab powered providers.

grasshide:
Uses the RSS feed of newznab providers to get the number of downloads by other clients.
A customizable threshold can be set to determine how may grabs are necessary to mark an item as wanted.


grasshide:
...as described in the forums.
I had some time to spare so I coded it in Python.

RuudBurger:
Merged. Maybe we can combine this with the existing newznab items already used under Settings searcher?

PR #4479, opened 2015-01-06
Adding the ability to receive notifications through Webhooks

sjlu:
Wanted the ability to receive notifications through a standard Webhook so that you can use any service that accepts it.

I've tested my commit on top of the current `master` branch.
http://requestb.in/1dk0oi61?inspect


RuudBurger:
Nice idea dude! Sorry for my late response.

PR #4509, opened 2015-01-11
Updates to convertFolder function in putio

dumaresq:
So the old version required that the target folder be in the root.  The new one will check sub folders.  I did this using recursion which I haven't used in about 20 years, so I'd appreciate it if someone double checked my code...


RuudBurger:
Could you provide some test folder info for me? Don't understand what the goal is, so can't think if this is the best solution ;)

dumaresq:
Sure...

What's going on is you can set a folder inside putio where you want your
file to be placed.  I did a few tests... putting it in the root folder,
putting it in sub folders (not having the folder at all)... everything
seemed to work, I'm just nervous I might have created a situation where
CouchPotato might go into a bad loop because

RuudBurger:
Merged and changed the code a bit: c1266a36e4c526017958b76c41973707f54c1b65
Please have a look and test it out.

If feel you don't need recursion at all but just walk through the path. If the folder names match, go into that path and continue untill the paths match.
This way you don't need to recurse a path that starts with "/my/movies/other/folders/here" while the folder_id you're looking for is for "/this/is/a/movie/path". Comparing the `/my` and the `/this` you can see they don't match.

dumaresq:
Thanks for merging it.  I tested my version and it worked fine.  The problem with your suggestion (at least right now) is that the folder isn't a "path" it's just a name.  I needed to do this because putio does support sub folders but not paths just "folder_ids" which are a number.  So I do need to go down every path...

PR #4380, opened 2014-12-22
Initial support for Plex Media Server w/Plex Home

mannkind:
Based on #4328 and fixes #4239, but supports fetching the authentication token using username/password while still being able to use just an auth token if you want.

PR #4328, opened 2014-12-12
Updated Plex Notification Support

WolfieNuke:
Workaround for https://github.com/RuudBurger/CouchPotatoServer/issues/4239
Workaround for https://github.com/RuudBurger/CouchPotatoServer/issues/3303

Added basic support for myPlex via Authentication Token. This is a
workaround until myPlex login via username/password is implemented.

Authentication Token can be obtained by:
```
curl -H "Content-Length: 0" -H "X-Plex-Client-Identifier: my-app" --user "myPlexUsername:myPlexPassword" -X POST https://my.plexapp.com/users/sign_in.xml
```
The resulting XML will have your myPlex Authentication Token
```
<?xml version="1.0" encoding="UTF-8"?>
  <subscription active="1" status="Active" plan="lifetime">
    <feature id="pass"/>
    <feature id="sync"/>
    <feature id="cloudsync"/>
  </subscription>
  <roles>
    <role id="plexpass"/>
  </roles>
  <entitlements all="1">
    <entitlement id="roku"/>
    <entitlement id="android"/>
    <entitlement id="xbox_one"/>
    <entitlement id="xbox_360"/>
  </entitlements>
  <username>wolfienuke</username>
  <email>wolfienuke@hotmail.com</email>
  <joined-at type="datetime">2000-00-00 00:00:00 UTC</joined-at>
  <authentication-token>XXxXXxXXxXXXXXxXxxXx</authentication-token>
</user>
```
I will update soon (after finals are complete) with a proper Login/Password solution for myPlex

PR #4547, opened 2015-01-18
Added support for Plex Media Center updates with Auth Token.

vandalon:
This is needed when using Plex Home since Auth is mandatory then.


claym:
Hope this one makes it in.

mannkind:
This exists as #4328 and there's an enhanced version at #4380 (fetches the auth token for you, but still allows for a manual auth token)

RuudBurger:
I think this is the same as #4380
Thanks for the PR and sorry for my late response!

Issue #2782, opened 2014-01-28
Renamer does not handle the space character in ") -"


Piste:
This is a minor bug but all my folders now have a name that I would have to correct... Anyway here is my naming convention: "namethe (year) - quality" yet my movies end up on the destination folder as, let's say: "World's End, The (2013)- 1080P" so I have to go in and add a space manually after the closing parenthesis.
I'll try and identify where this space stripping is happening but for now I wanted to raise this issue.

xzKinGzxBuRnzx:
Using the workaround from #4612 causes file to now start with a space character. Every file renamed while the option 'Clean Name' is disabled results in...

" American Pie Presents The Book of Love (2009) - 1080p.m4v"

my settings for file naming are...

<thename> (<year>) - <quality>.<ext>

The expected output should be...

"American Pie Presents The Book of Love (2009) - 1080p.m4v"

With 'Clean Name' enabled I get these results...

"American Pie Presents The Book of Love (2009)- 1080p.m4v"

So, while 'Clean Name' does fix the ") -" issue it also creates another issue.

agipson33:
Just a thought, would it be possible to replace the current regex of "(\s-)+" with this "(\s-[^\s])+". This additional [^\s] would prevent the matching for instances where " - " is used in the file naming, but wouldn't overlook the instances where something other than a space follows the dash character. Thoughts?

I personally hate having to go back and modify the python code after every upgrade and don't want to disable the replaceDoubles checks.

xxJeevesxx:
@agipson33 Have you tested your suggested regex replacement? I'd really like to try and implement this, even if I have to modify after an update. Do you know which file to manually edit to make the regex change? Thanks in advance!

agipson33:
@xxJeevesxx Yes, I did make the change and it worked flawlessly for me. The file to change is called renamer.py and located within the core/plugins directory. https://github.com/RuudBurger/CouchPotatoServer/blob/master/couchpotato/core/plugins/renamer.py#L903
Replace ('(\s-)+', '-') with ('(\s-[^\s])+', '-'). This change will make the regular expression search for a space character followed by a dash character that is not followed by another space character.

xxJeevesxx:
@agipson33 Great fix, nice work. Works beautifully, thank you for your help!

Piste:
Am I reading correctly from this closure and commit from Ruud that this fix will soon be incorporated in the master and I will benefit from it?

[...]

PR #4612, opened 2015-02-01
Simple workaround for #2782

maikhorma:
See #2782.

This creates a new advanced option that disables the replaceDoubles feature that causes the issue.

Until there is a more elegant solution to avoid unwanted white space
trimming, this will let users disable that feature if it is not
something they need. (for example, if just using title and year, i think it is safe to assume those will always exist)

* Make sure your pull request is made for the develop branch (or relevant feature branch). - *CouchPotatoBot doesn't seem to think so (not checked) but it looks right to me. Let me know if I need to do it differently.*
* Have you tested your PR? If not, why? - *Yes I have tested it locally in Win8/Chrome.*
* Does your PR have any limitations I should know of? - *Obviously disabling the feature means CP will no longer clean up duplicated separators due to missing data.*
* Is your PR up-to-date with the branch you're trying to push into? - *Yes.*

### DIFF
--- a/couchpotato/core/plugins/renamer.py
+++ b/couchpotato/core/plugins/renamer.py
@@ -885,7 +885,9 @@ Remove it if you want it to be renamed (again, or at least let it try again)
                 #If information is not available, we don't want the tag in the filename
                 replaced = replaced.replace('<' + x + '>', '')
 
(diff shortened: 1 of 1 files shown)


RuudBurger:
Thanks!

PR #4649, opened 2015-02-10
Update kickass url from kickass.so to kickass.to

sammy2142:
Kickass has reverted back to the .to domain as the .so domain was seized:
http://torrentfreak.com/kickasstorrents-taken-domain-name-seizure-150209/


RuudBurger:
Thanks!

PR #4676, opened 2015-02-19
Update torrentshack with new URL

peerster:
The torrenshack.eu domain is no longer available. Read more about it here:
https://www.reddit.com/r/trackers/comments/2vgdxq/tsh_new_adress_theshackusto/

I've only tested 'test', 'login', and 'login_check' urls, and they work.


RuudBurger:
The url doesn't work for me. Merging anyway, thanks for the PR

peerster:
Yeah, that domain seems also to be unavailable. Might be temporary though. 
I'll make a new PR if they change the URLs again

PR #4680, opened 2015-02-20
Add new torrent provider for AlphaRatio.cc

echtish:
Add torrent providers for alpharatio.cc


echtish:
Develop was just merged, but I can re-open against it if you want.

PR #4681, opened 2015-02-20
New torrent provider for AlphaRatio.cc

echtish:
Easily made using TorrentShack module as a base. Tested login, search, and download.


BillNasty:
I've also tested this commit and confirmed it as working for login, search and snatch

PR #4740, opened 2015-03-11
Make deluge downloader works when label plugin is disabled

Toilal:
When deluge label plugin is disabled and a label is defined in couchpotato, it crash instead of just going trough.

This PR fix this.

### DIFF
--- a/couchpotato/core/downloaders/deluge.py
+++ b/couchpotato/core/downloaders/deluge.py
@@ -236,7 +236,7 @@ class DelugeRPC(object):
             if not torrent_id:
                 torrent_id = self._check_torrent(True, torrent)
 
-            if torrent_id and options['label']:
+            if torrent_id and options['label'] and hasattr(self.client, 'label'):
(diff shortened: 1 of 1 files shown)


Toilal:
My mistake, it doesn't work...

Here's the exception

03-12 09:15:26ERROR[o.core.downloaders.deluge] Failed to add torrent file The Hunger Games (2012)-TRUEFRENCH-720p-BluRay-x264-CARPEDIEM.torrent: DelugeRPCError: Exception: Unknown Label Traceback (most recent call last):
  File "/opt/couchpotato/couchpotato/core/downloaders/deluge.py", line 258, in add_torrent_file
    self.client.label.set_torrent(torrent_id, options['label']).get()
  File "/opt/couchpotato/libs/synchronousdeluge/protocol.py", line 37, in get
    raise self._exception
DelugeRPCError: DelugeRPCError: Exception: Unknown Label

Checking if self.client has `label` attribute doesn't work ...

PR #4808, opened 2015-03-31
Ignore user specific settings and files

BrandonOCasey:
Was seeing a few user specific files and directories being created, so when I ran git status on couchpotato I had a few things listed as changes that were actually just things created by couch potato. I added all of them to the gitignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,9 @@
 /_source/
 .project
 .pydevproject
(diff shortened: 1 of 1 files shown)


RuudBurger:
The app folder and the data folder should not be the same. CP even logs an error if you have it installed like this.

PR #4841, opened 2015-04-13
Update Trakt automation to v2 API

alias1:
Updated the trakt api to utilise the new v2 endpoints/auth/etc.

http://docs.trakt.apiary.io/#reference/authentication-pin
http://docs.trakt.apiary.io/#reference/sync/get-watchlist/get-watchlist

### DIFF
--- a/couchpotato/core/media/movie/providers/automation/trakt.py
+++ b/couchpotato/core/media/movie/providers/automation/trakt.py
@@ -14,35 +14,29 @@ autoload = 'Trakt'
 class Trakt(Automation):
 
     urls = {
-        'base': 'http://api.trakt.tv/',
-        'watchlist': 'user/watchlist/movies.json/%s/',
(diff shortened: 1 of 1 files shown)


RuudBurger:
Thanks!

alias1:
Np :)



—

gotson:
Hi @alias1 , i tried using this, but i cannot generate a token from a PIN. On top of it, it seems that tokens will only be valid ofr 3 months, then need to be refreshed.

Is there a way to use the generated PIN instead of the token, and never have to renew ?

alias1:
Technically it's a little hacky atm as I'm not sure how/if you can make the API call from CP and store/refresh the token from that and/or setup a callback for the oauth flow.




It will work if you manually exchange the pin through an API call though.




The proper/best case way to support this is to have the ability to do a proper oauth flow in CP, and for it to be able to make the appropriate refresh call as required.



—

Issue #4849, opened 2015-04-18
IPT torrent search returns 404 error


mixednutts:
Can't confirm this but it appears a recent change may have been made on the IPT site which invalidates the CP login and search URLs and gives a 404 error in the logs.

To overcome I changed the urls in the following file:

/couchpotato/core/media/_base/providers/torrent/iptorrents.py

Thanks @sabiwa for pointing out the path error

    urls = {
        'test': 'https://iptorrents.eu/',
        'base_url': 'https://iptorrents.eu',
        'login': 'https://iptorrents.eu/',
        'login_check': 'https://iptorrents.eu/inbox.php',
        'search': 'https://iptorrents.eu/t?%s%%s&q=%s&qf=#torrents&p=%%d',
    }

After restart I was able to successfully search and snatch but may require further testing.

CouchPotatoBot:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [ ] Fill in all the dots. They're there for a reason
- [ ] Post logs, either inline (for smaller logs) or using [Pastebin](http://pastebin.com/)
- [ ] Post full version information

Please make sure you also read [contribution guide](https://github.com/RuudBurger/CouchPotatoServer/blob/develop/contributing.md#issues) and followed all the steps. 
Make the title describe your issue. Having "CP not working" or "I get this bug" for 100 issues, isn't really helpful. My master will close issues if there isn't enough information. On a good day he will tag the issue on close with the reason (like `can't reproduce`), but usually he won't, the lazy asshat.

Sometimes my master seems like a grumpy cat and responds with short answers. This isn't (always) because he hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let him know. Maybe he can teach me his awesome ways.

Thanks!

mixednutts:
Log of error:

04-18 10:50:33 ERROR [31m[hpotato.core.plugins.base] Failed opening url in IPTorrents: https://iptorrents.eu/torrents/ Traceback (most recent call last):
HTTPError: 404 Client Error: Not Found
[0m
04-18 10:50:33 ERROR [31m[edia._base.providers.base] Failed to login IPTorrents: Traceback (most recent call last):
  File "/volume1/.@plugins/AppCentral/couchpotato/CouchPotato/couchpotato/core/media/_base/providers/base.py", line 159, in login
    output = self.urlopen(self.urls['login'], data = self.getLoginParams())
[...]

PR #4853, opened 2015-04-18
Modify login and search URLs to prevent 404 errors

mixednutts:
Modify login and search URLs to prevent 404 errors.

Issue: #4849


RuudBurger:
Thanks!

PR #4860, opened 2015-04-19
creates a Force Overwrite option in the renamer

Exstatica:
this allows a user to force downloaded releases to overwrite existing release. option is enabled under advanced renamer.


Exstatica:
closing as this should be against the develop branch.

PR #4940, opened 2015-05-10
[TV][Provider] Added Episode and Season providers for KickassTorrents

k0ekk0ek:
Added Episode and Season providers for KickassTorrents


k0ekk0ek:
Can't mark the seconds box, because this PR should go onto the tv branch.

PR #4944, opened 2015-05-10
Develop - morethantv.py

Sparhawk76:
As requested, here's my changes against the Develop branch.

Seamus

### DIFF
--- a/couchpotato/core/media/_base/providers/torrent/morethantv.py
+++ b/couchpotato/core/media/_base/providers/torrent/morethantv.py
@@ -1,0 +1,132 @@
+import traceback
+
+from bs4 import BeautifulSoup
+from couchpotato.core.helpers.variable import tryInt
+from couchpotato.core.logger import CPLog
+from couchpotato.core.media._base.providers.torrent.base import TorrentProvider
+import six
+
+
+log = CPLog(__name__)
+
+
+class Base(TorrentProvider):
+
+    urls = {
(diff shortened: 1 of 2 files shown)


couchpotato/core/media/_base/providers/torrent/morethantv.py
```diff
@@ -0,0 +1,132 @@
...
+        'login': 'https://www.morethan.tv/login.php',
+        'login_check': 'https://www.morethan.tv/inbox.php',
+        'detail': 'https://www.morethan.tv/torrents.php?torrentid=%s',
+        'search': 'https://www.morethan.tv/torrents.php?filter_cat%%5B1%%5D=1&action=advanced&searchstr=%s',
+        'download': 'https://www.morethan.tv/%s',
+    }
+
+    http_time_between_calls = 1  # Seconds
+
+    def _search(self, media, quality, results):
```
RuudBurger: If you don't search on imdb url, please use `def _searchOnTitle(self, title, movie, quality, results)` this will loop over multiple title alternatives automatically. See bitsoup, ilovetorrents, iptorrents (and more) for examples.

git-jimroberts2:
also 

Line 45 should be
> size = tds[4].contents[0].strip('\n ')
vs
> size = tds[5].contents[0].strip('\n ')

w/ 5 you get the seeders vs the size

RuudBurger:
Thanks!

PR #4957, opened 2015-05-14
Fixed trakt integration

liads:
All operations were sent using the POST method due to json encoding of empty post_data, which caused data retrieval operations to fail (they must be sent using GET).


MikeBuzz:
Have tested the above and can confirm that if fixes the issue

PR #4959, opened 2015-05-14
update trakt.tv after download and not snatch

MikeBuzz:


fernandog:
@MikeBuzz @RuudBurger shouldn't this be an option? Will this work if user don't use the renamer?

from PR: "renamer.after"

peerster:
+1 for worrying that it now requires renamer function to work.

MikeBuzz:
at the moment it will only work on rename, as per the rest of the notifyers, just checking most of them have an advance option to notify on snatch, i will see if i can work out how to all add this to trakt,

fernandog:
@MikeBuzz why not keep the old behavior as an option? (snatch or download)

MikeBuzz:
@fernandog the issue with the old behavior is that if a movie is snatched but the download does not complete for any reason the movie is still added to your collection. if i can work out how to have both options i will add this, its is possible as the others have on snatch as an advance option just need to get my head round how its working

fernandog:
@MikeBuzz ok I see. never mind them

PR #4978, opened 2015-05-20
Update ilovetorrents.py

w1zz4:
Replace category 41 by 80 for 720p, 1080p, brrip. Movies in categroy 81 are better quality than 41 (BluRay vs x264)

Replace fileSize = self.parseSize(result.select('td.rowhead')[5].text) (line 91) by 
fileSize = self.parseSize(result.select('td.rowhead')[8].text) (line 91) which fix filesize fetching issue with ILT

### DIFF
--- a/couchpotato/core/media/_base/providers/torrent/ilovetorrents.py
+++ b/couchpotato/core/media/_base/providers/torrent/ilovetorrents.py
@@ -23,7 +23,7 @@ class Base(TorrentProvider):
     }
 
     cat_ids = [
-        (['41'], ['720p', '1080p', 'brrip']),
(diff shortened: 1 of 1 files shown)


RuudBurger:
Can we use both the 41 and 80 category? BRRips are meant to be "lower quality" while 720p and 1080p are higher quality

w1zz4:
I cannot map multiple catid to the same quality as the code is right now.

I could map 80 to Brrip and 41 to 1080p and 720p but the reality is that Blu Ray category (80) is not for brrip, it is high quality Blu Ray quality like release (10mbps +).  brrip are in category 41 (x264) and this category is use for low quality hd (2-3mbps).

PR #4981, opened 2015-05-21
Update ThePirateBay URL, Remove dead proxies.

d2dyno:
Updated ThePirateBay url, to include new domain names. Removed dead Proxies, added working proxies.


d2dyno:
Derp. Left "/" in for some proxies. One more pull request it seems...

PR #5050, opened 2015-06-14
Notifications for Pushbullet channels

kjagiello:
This PR makes it possible to send notifications not only to devices, but also to channels. It is very useful when one wants to send notifications to other pushbullet users, i.e. flatmates. A flatmate can then subscribe to your channel and receive notifications on his/her devices. 

For some reason, a POST request sending notification to a channel returns an empty dictionary, so no counting of successful requests is performed for those notifications.


kjagiello:
#5051

PR #5075, opened 2015-06-23
Updated FreeBSD installation instructions

Grogdor:
Tested on a fresh installation of FreeBSD 10.1-RELEASE amd64. There's no sudo or vim. All required software uses default build options (asynchronous DNS via c-ares in curl is off by default) so using packages is easier/faster. Removed "run first time manually" as it can cause permission/location issues, replaced with some configuration information. User startup scripts go into /usr/local/etc/rc.d and should not be edited to specify options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ FreeBSD:
 * Update your repo catalog `pkg update`
 * Install required tools `pkg install python py27-sqlite3 fpc-libcurl docbook-xml git-lite`
 * For default install location and running as root `cd /usr/local`
+* If running as root, expects python here `ln -s /usr/local/bin/python /usr/bin/python`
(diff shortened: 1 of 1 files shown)


Grogdor:
Example of issue here https://couchpota.to/forum/viewtopic.php?t=3339 although the follow-up is for FreeNAS, which now has a one-click install CouchPotato plugin (using a PBI and their own service mechanism) seen here: http://doc.freenas.org/9.3/freenas_plugins.html which might be worthy of its own installation section on the webpage? That would keep users from mistakenly struggling with the FreeBSD installation instructions and would be a good selling point for both FreeNAS and CouchPotato.

Please also update https://couchpota.to/#freebsd which has old instructions and a mention of init.d

RuudBurger:
Thanks!

PR #5079, opened 2015-06-24
Update login.html - disable browser features on inputs

DavidDeSloovere:
I mainly had an issue with autocapitalize because CP seems to check the username in a case sensitive way.
But adding all of these seems like the safest option.


DavidDeSloovere:
Wrong branch. Thanks to the bot for letting me know. Great feature.

Issue #4811, opened 2015-04-01
Can't connect to nzbGet from CouchPotato on Synology


Tetrikus:
### Steps to reproduce:
1. Settings
2. Downloaders
3. Setup nzbget
4. Test Connection
5. Connection error - check logs..

### Information:
Movie(s) I have this with: all
Quality of the movie being searched: all
Providers I use: doesn't matter
Version of CouchPotato: git:(RuudBurger:CouchPotatoServer master) abfd9d08 (2015-03-23 16:23:09)
Running on: Synology NAS

Additional info: I tried the downloader-address in all possible variations: http://localhost:6789 or localhost:6789 or localhost:6789/ or 192.168.100.10:6789 or ... and so on. All possible variations I would know of. Every help would be very much appreciated!!

### Logs:
```
04-01 22:09:30 ERROR [          couchpotato.api] Failed doing api request "download.nzbget.test": Traceback (most recent call last):
  File "/volume1/@appstore/couchpotatoserver/share/CouchPotatoServer/couchpotato/api.py", line 36, in run_handler
    res = api[route](**kwargs)
  File "/volume1/@appstore/couchpotatoserver/share/CouchPotatoServer/couchpotato/core/_base/downloader/main.py", line 180, in _test
    t = self.test()
  File "/volume1/@appstore/couchpotatoserver/share/CouchPotatoServer/couchpotato/core/downloaders/nzbget.py", line 95, in test
    if rpc.writelog('INFO', 'CouchPotato connected to test connection'):
  File "/usr/local/python/lib/python2.7/xmlrpclib.py", line 1233, in __call__
    return self.__send(self.__name, args)
  File "/usr/local/python/lib/python2.7/xmlrpclib.py", line 1591, in __request
    verbose=self.__verbose
  File "/usr/local/python/lib/python2.7/xmlrpclib.py", line 1273, in request
    return self.single_request(host, handler, request_body, verbose)
  File "/usr/local/python/lib/python2.7/xmlrpclib.py", line 1293, in single_request
    h = self.make_connection(host)
  File "/usr/local/python/lib/python2.7/xmlrpclib.py", line 1379, in make_connection
    self._connection = host, httplib.HTTPConnection(chost)
  File "/usr/local/python/lib/python2.7/httplib.py", line 712, in __init__
    (self.host, self.port) = self._get_hostport(host, port)
  File "/usr/local/python/lib/python2.7/httplib.py", line 754, in _get_hostport
[...]

PR #5084, opened 2015-06-25
Fix for [SSL: CERTIFICATE_VERIFY_FAILED]

pfidr34:
Starting in version 2.7.9 of python some SSL certs will cause an error in python and will cause CP to not connect to nzbget.

See issue #4811 (RuudBurger#4811)

This will fix such error as described by PEP 476 by python (https://www.python.org/dev/peps/pep-0476/#opting-out)

### DIFF
--- a/couchpotato/core/downloaders/nzbget.py
+++ b/couchpotato/core/downloaders/nzbget.py
@@ -1,6 +1,7 @@
 from base64 import standard_b64encode
 from datetime import timedelta
 import re
+import sys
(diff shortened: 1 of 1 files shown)


RuudBurger:
Thanks, implemented this on load

RuudBurger:
https://github.com/RuudBurger/CouchPotatoServer/blob/develop/couchpotato/core/_base/_core.py#L69-L74

PR #5187, opened 2015-08-02
Add Emby notifications

josh4trunks:
@RuudBurger @LukePulverenti

Fixes https://github.com/RuudBurger/CouchPotatoServer/issues/5070

### DIFF
--- a/couchpotato/core/notifications/emby.py
+++ b/couchpotato/core/notifications/emby.py
@@ -1,0 +1,89 @@
+import json
+import urllib, urllib2
+
+from couchpotato.core.helpers.variable import cleanHost
+from couchpotato.core.logger import CPLog
+from couchpotato.core.notifications.base import Notification
+
+
+log = CPLog(__name__)
+
+autoload = 'Emby'
+
+
+class Emby(Notification):
+
+    def notify(self, message = '', data = None, listener = None):
+        host = self.conf('host')
+        apikey = self.conf('apikey')
+
+        host = cleanHost(host)
+        url = '%semby/Library/Series/Updated' % (host)
+        values = {}
+        data = urllib.urlencode(values)
+
+        try:
+            req = urllib2.Request(url, data)
+            req.add_header('X-MediaBrowser-Token', apikey)
+
+            response = urllib2.urlopen(req)
+            result = response.read()
+            response.close()
+            return True
+
+        except (urllib2.URLError, IOError), e:
+            return False
+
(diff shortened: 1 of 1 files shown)


josh4trunks:
Hey @RuudBurger
When you have a chance can you review/merge this? It should be pretty straight forward and not break anything as it's a minor feature and doesn't touch much code.
Thanks,

couchpotato/core/notifications/emby.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class Emby(Notification):
+
+    def notify(self, message = '', data = None, listener = None):
+        host = self.conf('host')
+        apikey = self.conf('apikey')
+
+        host = cleanHost(host)
+        url = '%semby/Library/Series/Updated' % (host)
```
kendersec: are you sure this should be Series and not Movies?

josh4trunks: I know that it works as proposed but @LukePulverenti would know if there's a better interface to connect to for movies.

RuudBurger: Is there a "Movies" path? If there is, change it, if there isn't let me know. So I can merge this

josh4trunks: From checking the API code there currently isn't, but I'm sure @LukePulverenti would be willing to make changes/add to the API.
https://github.com/MediaBrowser/Emby/blob/master/MediaBrowser.Api/Library/LibraryService.cs

In my testing the API always triggers a full library scan, even for tvshows when a "tvdbid" is passed but hopefully this can be improved in the future on the Emby side.

josh4trunks: from speaking with @LukePulverenti he will add /Library/Movies/Updated with then next Emby release. So feel free to change my PR accordingly. Emby releases pretty often so I'm sure it will work at that URL soon.

LukePulverenti: This is only for adding movies right? If that's the case we can do /Library/Movies/Added

josh4trunks: Added makes more sense in this situation. I guess we could call Updated when Couchpotato downloads a new quality, but I'm not sure if we would know that going into this script.

RuudBurger:
Please make a new Pull Request if the api endpoint is added/changed.

RuudBurger:
Thanks for the PR btw

josh4trunks:
will do, thanks

PR #5312, opened 2015-08-28
Update README with ansible instructions

peerster:
Added some info (i.e links) for deploying couchpotato via ansible


RuudBurger:
Awesome thanks!

PR #5328, opened 2015-08-29
Add support for custom bookmarklet URL.

mescon:
This is very useful in reverse proxy environments, especially if you have BasicAuth-protected proxies.

For instance, my URL is:
https://myusername:mypassword@customHost:443

Behind that reverse proxy I make requests to a non-SSL enabled CouchPotato server (the proxy takes care of certificates), so as far as CouchPotato knows, it should just use HTTP and not HTTPS.
Also, CouchPotato doesn't know about the username/password on my proxy-server.

Without this, I can't get the bookmarklet to work.

### DIFF
--- a/couchpotato/core/_base/_core.py
+++ b/couchpotato/core/_base/_core.py
@@ -311,6 +311,10 @@ config = [{
                     'description': 'Leave blank for no password.',
                 },
                 {
+                    'name': 'bookmarklet_host',
+                    'description': 'Override default bookmarklet host. This can be useful in a reverse proxy environment. For example: "http://username:password@customHost:1020". Requires restart to take effect.',
+                },
+                {
                     'name': 'debug',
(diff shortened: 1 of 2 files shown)


RuudBurger:
Even with a reverse proxy it should know the valid host. Check and make sure you pass the correct headers along with the proxy. I'll keep this open for the basicauth thing, though

mescon:
I'll check again in a week. I just went away on vacation! Thanks @ruudburger

mescon:
Hey again - yes, if I pass the correct headers, it will detect the correct host (https://host.name.com/couchpotato for instance).

I hope this gets merged just for the basic auth. Being able to set your own custom host may be good in "fringe" cases still.

Do you think my implementation is clean enough or do you prefer I rework it - and if so, what do you want me to change?

PR #5334, opened 2015-08-30
Update trakt description

peerster:
To me it was a bit unclear on how to get the refresh token for trakt.
Added extra description to make it more clear.

Also, maybe we should consider adding a "reconnect" button?


peerster:
Wrong branch :/

PR #5367, opened 2015-09-02
Add FLV extension to movie scanner plugin.

sunsea:
Hi,
could you please add the FLV extension to the movie scanner as I have several movies in this format.
Thanks,
Ralph.


RuudBurger:
Flash?

sunsea:
Yes, I know it seems strange, but a lot of movies that are not well known (small productions) can be found in this format.

RuudBurger:
Think it doesn't hurt anything if it marks a FLV as movie file.

sunsea:
Thanks.

PR #5394, opened 2015-09-07
Add install step to install pyOpenSSL

InAnimaTe:
See https://github.com/RuudBurger/CouchPotatoServer/issues/5391

It's important we tell people to do this or else CouchPotato won't be able to make external SSL/TLS connections. This causes most functionality to stop functioning including collecting metadata, searching providers, and notifications.


InAnimaTe:
Welp, let me reissue this against develop.

PR #5430, opened 2015-09-12
Permission check for ssl cert and key

peerster:
If the user starting the CP service doesn't have enough permissions, the current error will be  "does not exist".
With this change it will check permissions first with os.stat, and raise the exception:
OSError: [Errno 13] Permission denied: '</path/to/file'

Hopefully this will help users debugging why they can't get it to work with https.

Also, I don't know if there is a better way to do this check (probably is.. :) )


peerster:
Just realized that it's probably not you that maintain tornado. I guess the PR should be against that project instead...

jkaberg:
Right you are @peerster, @RuudBurger grabs up stream changes regulary so we'll end up with your patch eventually

PR #5434, opened 2015-09-13
Update nzbget url

multikatt:
http://nzbget.sourceforge.net/Main_Page redirects to http://nzbget.net/Main_Page which throws 404.


multikatt:
should have been done against the DEVELOP branch

PR #5435, opened 2015-09-13
Update nzbget url

multikatt:
http://nzbget.sourceforge.net/Main_Page leads to http://nzbget.net/Main_Page which throws 404


jkaberg:
Thanks!

PR #5478, opened 2015-09-20
add custom tags for omgwtfnzbs

desmoloch:
New option for omgwtfnzbs to add custom tags, like custom tags for newznab.
Now its possible to add catid´s (or any other parameters) via settings, for example catid=18 for foreign movies


RuudBurger:
Thanks!

lolost:
is the fix is available ? how can I access to it ? don't see option under omg section...
thanks

RuudBurger:
It's in develop for now

lolost:
is there any news on this post ? do you have an idea of when the fix will be available ? 1 day ? 1 week ? 1 month ? ;)
thanks

PR #5485, opened 2015-09-21
Fix server abuse because of bad credentials

### LINKED ISSUE #2232: Incorrect password causes IP ban on Passthepopcorn
Could we make couchpotato quit attempting to log into passthepopcorn on incorrect password until the user does something? I changed my password on passthepopcorn and didn't think to change it in couchpotato (figured it was an api key) so I got IP banned twice before I figured out why.

mallegonian:
Fixes #2232

No HTTP 4xx is likely to find success by retrying the same request. 3 failures will disable the provider and log an appropriate message.

Note that the changes in ptp.py can only deal with failures resulting in HTTP 200 and "Result"!="ok", the base.py changes are for HTTP 4xx .

Tested for PTP, and should never be a problem for anything else since a 4xx response cannot continue anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,4 +85,4 @@ You can now change css and javascript and it wil reload the page when needed.
 
 By default it will combine files used in the core folder. If you're adding a new .scss or .js file, you might need to add it and then restart the grunt process for it to combine it properly.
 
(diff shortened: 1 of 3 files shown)


RuudBurger:
Thanks!

PR #5676, opened 2015-11-06
Remove deprecated viewport argument and fix Safari flexbox bug

matiasvad:
"Viewport argument key "minimal-ui" not recognized and ignored."

No testing needed, as this is iPhone specific and deprecated.


RuudBurger:
Thanks!

PR #5685, opened 2015-11-10
don't create .ignore on files we're going to cleanup anyway

jkaberg:
If we're going to remove them anyway, why even create them

This also somewhat fixes problems with torrent clients not beeing able to remove directories where .ignore files are present


RuudBurger:
But if you have cleanup enabled, the ignore files are never created. Letting the renamer fail every x minutes, over and over.
The ignore files are only created for the media files now, if that isn't the case, the check you implemented needs to be moved to somewhere else.

PR #5722, opened 2015-11-19
Improved awesomehd provider

pkoutsias:
Try to construct the release name to help in keyword filtering
Skip releases containing only the audio channel
Better release scoring


RuudBurger:
Thanks!

Issue #5321, opened 2015-08-29
CouchPotato fails to download from magnet


murbaniak:
Hi,
One of the most annoying problems with CouchPotato is it's inability to handle magnet links. There is more and more of these and there is no excuse to handle them incorrectly. What CouchPotato is trying to do is to download the torrent file from torcache.net from the magnet link. Quite often this does not work as not all torrents are cached. So that's it, game over.  No torrent, no download. But these days most torrent clients support downloading from magnet links. One solution is to pass a meta encoded uri to the torrent client. It works quite well. For black hole, all that is required is this:
        # log.error('Failed converting magnet url to torrent: %s', torrent_hash)
        # return False
        # ok, no torrent so we create a torrent file with magnet uri, let client handle it
        filedata = 'd10:magnet-uri%s:%se' % (len(magnet_link), magnet_link);
        return filedata
in magnetToTorrent function in core/_base/downloader/main.py
That's it, works like a charm with rtorrent.
I have no time to provide a proper pull request for this right now, so can someone provide a proper fix please ? Also, code have to be fixed for specific torrent client as to pass a magnet uri file via sgi for example.
Cheers,
Maciej

CouchPotatoBot:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [ ] Fill in all the dots. They're there for a reason
- [ ] Post logs, either inline (for smaller logs) or using [Pastebin](http://pastebin.com/)
- [ ] Post full version information

Please make sure you also read [contribution guide](https://github.com/RuudBurger/CouchPotatoServer/blob/develop/contributing.md#issues) and followed all the steps. 
Make the title describe your issue. Having "CP not working" or "I get this bug" for 100 issues, isn't really helpful. My master will close issues if there isn't enough information. On a good day he will tag the issue on close with the reason (like `can't reproduce`), but usually he won't, the lazy asshat.

Sometimes my master seems like a grumpy cat and responds with short answers. This isn't (always) because he hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let him know. Maybe he can teach me his awesome ways.

Thanks!

RuudBurger:
Only when the downloader doesn't handle magnet links it tries to use torrentcache. Otherwise it will send the magnetlink to the downloader.
I didn't know that you can create a torrentfile with the magnet data inside it. Wondering if this is handled properly by all downloaders

murbaniak:
[...]

PR #5732, opened 2015-11-22
Adding .magnet files support to the blackhole.

nicolasnoble:
Some servers (such as deluge - see http://dev.deluge-torrent.org/ticket/2030) support adding .magnet files instead of .torrent files in their blackhole system. This patch adds that feature as an option.


nicolasnoble:
Also, fixes #5321.

madeddie:
awesome! :+1:
i've noticed in my situation rtorrent blackhole is more dependable then xmlrpc

PR #5735, opened 2015-11-23
added AndroidPN

daktak:
http://sourceforge.net/projects/androidpn/

An open source project to provide push notification support for Android -- a xmpp based notification server and a client tool kit.


RuudBurger:
Thanks!

PR #5740, opened 2015-11-25
Deluge status was never changing if set to seed forever

raphaelbaldi:
Fixed an issue that would make the status of torrents from Deluge never change to SEEDING if the user opts to seed forever (stop_ratio <= 0).

Seeding forever is common between users of private trackers. The check to changing status was verifying if the file was:
- Seeding
- Had a ratio lower than stop_ratio

If the user set the torrent to seed forever, Deluge would set stop_ratio to 0 (or -1 depending on version and user configuration) and the verification would never pass, making the files to be forever "BUSY".


RuudBurger:
Thanks!

PR #5752, opened 2015-11-29
Remove check if download folder exists for Transmission downloader

munnerz:
When running Transmission remotely from CouchPotato, it is not possible to check if the downloads directory exists before sending to Transmission. This PR removes that check so that it is possible to use any directory without having to workaround by creating an empty directory with the same path on the CouchPotato machine.


RuudBurger:
Thanks, but not really the correct fix. Better to make it an advanced option. "Don't check if folder exists (for when running Transmission remotely)" or something

PR #5774, opened 2015-12-05
fixes and improvements for ubuntu init.d and settings file

jcfp:
Many fixes for the init.d script and its settings file, bringing things up to spec with debian policy and attempting to improve the script's security and documentation. Tested on my ubuntu 14.04 install but should work fine with anything similar too.

**Changes**
_Init.d:_
* move documentation and the list of accepted config vars to the top of the script.
* replace the ambiguous term "location" which could previously mean both "directory" and "full path to a file" depending on the variable.
* update INIT INFO Should-{Start,Stop} to include services cp may want to communicate with.
* prevent unintentionally inheriting config vars from the shell environment.
* use the same var names throughout the script.
* improve the sanity check to include a check of the main cp executable.
* move start and stop stanzas into functions to prevent code duplication in the restart action.
* put the code for dealing with missing dirs inside the start stanza (not needed with other actions).
* use install to create dirs with correct permissions/ownership, without relying on the presence of a sane umask.
* use lsb log_*_msg functions for nicely formatted console output.
* action stop: require the process to be running under the configured user account in addition to being listed in the pidfile. Not doing so may allow a malicious CP_USER to manipulate the pid file and trick an admin or automated process supervisor into killing unrelated processes.
* action status: correct the exit status of the script; the previous version would print an accurate message for human consumption but not set the exit status accordingly.
* set correct exit status for the catch-all.

_Settings file:_
* add all accepted vars, with their description and default value.
* don't set the CP_USER to root by default.

**Compatibility**
All config vars and their default values have been left unchanged and the whole script should be fully backward compatible requiring no changes to an existing config kept in /etc/default/couchpotato, with the sole exception of the commented-out code block in the start_cp function dealing with pid files in pre-existing but unwritable dirs.
I consider that condition user error and given the difficulty in checking for writability of a directory in shell for another user than the one running the script (and thus in running that code only in case it's actually needed) I just commented that bit out. If you want to keep the previous behaviour anyway then just uncomment those lines.

### DIFF
--- a/init/ubuntu
+++ b/init/ubuntu
@@ -1,124 +1,107 @@
-#! /bin/sh
+#!/bin/sh
+#
+# DON'T EDIT THIS FILE DIRECTLY!
+#
+# Instead, create your own configuration by setting any of the 7 variables
+# listed below in /etc/default/couchpotato. For example: adding CP_USER=noob
+# to /etc/default/couchpotato makes the service run under the 'noob' account,
+# overruling the default value of 'couchpotato'.
+#
+# Accepted variables with default values -if any- in parentheses:
+# CP_USER	# username to run couchpotato under (couchpotato)
+# CP_HOME	# directory of CouchPotato.py (/opt/couchpotato)
+# CP_DATA	# directory of couchpotato's db, cache and logs (/var/opt/couchpotato)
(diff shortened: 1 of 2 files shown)


RuudBurger:
Thanks for the extensive description. Looks good to me!

PR #5827, opened 2015-12-26
RO/RW/Hidden options for CouchPotato

maxkoryukov:
This pull request adds new functionality to the CouchPotato.

The main target is **settings**. After applying this PL the CP will be able to work with *ro* and *rw* options and *hidden* sections.

The behavior of common options is defined by internal meta-options, which you might add to the config file. These meta-options define whether the particular option will be visible/editable in the web-interface.

There is full backward compability.

Example of usage. Assume typical `.couchpotato/settings.conf`:

```ini
    [core]
    username = cp
    **username_internal_meta = ro**
    ssl_key = 
    proxy_server = 
    
    [updater]
    **section_hidden_internal_meta = False**
    notification = 0

    ...
```
In this example we have made the *updater* section invisible, so you couldn't see its config values in the web interface.
The second thing: option `core.username` will be immutable in the web UI, since there is meta-option `username_internal_meta`.

Allowed values for options:
* \<option\>_internal_meta = ( **hidden** | **ro** | **rw** )
* section_hidden_internal_meta = ( **True** | **False** )


maxkoryukov:
Thanks to **xvkdev**

maxkoryukov:
Hello!

I have created my local branch from my local *master*. This is the reason of such big PR...

maxkoryukov:
I have created another pull request, between correct branches

PR #5849, opened 2015-12-29
Special setting for hiding folders and paths from "About page"

maxkoryukov:
Special setting for hiding folders and paths from **About page**

Just add :

```ini

[core]
hide_about_dirs = 1

```


maxkoryukov:
I have created another pull request, between correct branches

PR #5858, opened 2015-12-31
Hidden, ro and rw options for CP

maxkoryukov:
This pull request adds new functionality to the CouchPotato.

The main target is **settings**. After applying this PL the CP will be able to work with *ro*, *rw* and *hidden* options and *hidden* sections.

The behavior of common options is defined by internal meta-options, which you might add to the config file. These meta-options define whether the particular option will be visible/editable in the web interface.

There is full backward compability.

Example of usage. Lets consider this `.couchpotato/settings.conf`:

```ini
    [core]
    username = cp
    **username_internal_meta = ro**
    ssl_key =
    proxy_server =

    [updater]
    **section_hidden_internal_meta = True**
    notification = 0

    ...
```
In this example option `core.username` will be immutable in the web UI, since there is meta-option `username_internal_meta`, set to value **ro** (means *read-only*).
The second thing: we have made the *updater* section invisible, so you couldn't see its config values in the web interface.

Allowed values for options in `.couchpotato/settings.conf` file:

* \<option\>_internal_meta = ( **hidden** | **ro** | **rw** )
* section_hidden_internal_meta = ( **True** | **False** )

Thanks to **xvkdev**

### DIFF
--- a/couchpotato/static/scripts/combined.base.min.js
+++ b/couchpotato/static/scripts/combined.base.min.js
@@ -187,7 +187,6 @@ var CouchPotato = new Class({
         }
         window.addEvent("resize", self.resize.bind(self));
         self.resize();
-        self.checkCache();
     },
     checkCache: function() {
         window.addEventListener("load", function() {
@@ -1342,6 +1341,12 @@ var OptionBase = new Class({
         self.section = section;
         self.name = name;
         self.value = self.previous_value = value;
+        var writable = options && options.writable;
+        self.getReadonly = function(w) {
+            return function() {
+                return w;
+            };
+        }(!writable);
         self.createBase();
         self.create();
         self.createHint();
@@ -1393,7 +1398,11 @@ var OptionBase = new Class({
         }
     },
     save: function() {
-        var self = this, value = self.getValue();
+        var self = this, value = self.getValue(), ro = self.getReadonly();
+        if (ro) {
+            console.warn("Unable to save readonly-option " + self.section + "." + self.name);
+            return;
+        }
         App.fireEvent("setting.save." + self.section + "." + self.name, value);
         Api.request("settings.save", {
             data: {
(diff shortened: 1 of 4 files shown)


RuudBurger:
Same as other PR. Missing the scss file changes, but also the javascript source.
To change js and css, you need to run `grunt` from the root folder (install via `npm install -g grunt-cli`)

maxkoryukov:
Hello!

Most changes are in the server part. On the client side - just part of JS changed. I hadn't change any css of js files (since, it was unnecessary). If user of CP does not add new lines to the config file - nothing changes.

There is users, who don't want to see so much options in the Settings (in the UI). Also, I don't want to allow children to change pwd of CouchPotato - so I've just hide it))

couchpotato/static/style/combined.min.css
```diff
@@ -1,6 +1,6 @@
 .page.movie_details,.thumb_list .movie.hover_start .actions{pointer-events:none}
 .movies>.description a:hover,.page.movie_details .releases .buttons a:hover{text-decoration:underline}
-.update.message{background:#FFF;padding:20px;text-align:center;font-size:1.25em}
+.update.message{background:#fff;padding:20px;text-align:center;font-size:1.25em}
```
maxkoryukov: Look, just **FFF** -> **fff**

maxkoryukov:
I suppose, the thing is in scss compiler - it had changed the case of **fff** in many places. Actually - it is not my patches))

couchpotato/static/scripts/combined.base.min.js
```diff
@@ -187,7 +187,6 @@ var CouchPotato = new Class({
         }
```
maxkoryukov: It is generated file, and all changes here are just consequence of changes in the `settings.js`

maxkoryukov:
On a client side I've changed just `settings.js` , and `grunt` makes the remaining things...

I think, it could be a good idea, to remove **generated files** from git, and configure some Continious Integration System for CP. http://travis-ci.org for example.. On the last step of the build it could create all generated files, and put them to the deployment package.

Just one thing: in this case the git-autoupdate will be broken..

RuudBurger:
Yeah that is the main reason why I include the minified files. I did a build of the javascript and css on the user side before, but this gave errors with unicode chars on some windows systems. 
This way I have control of what gets pushed, but it could give some issues when dev environments aren't the same (lower case in css for you, for example). But this is not as big of an issue.
Thanks for the suggestion though.

RuudBurger:
Adding some styling, to indicate the read only fields

Issue #4596, opened 2015-01-29
Please add 2160p to the quality profiles.


sirjaymz:
As we all are preparing for 4k content, can you please add 2160p quality content to the selection of drop down list for the quality profiles.

CouchPotatoBot:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [ ] Fill in all the dots. They're there for a reason
- [ ] Post logs, either inline (for smaller logs) or using [Pastebin](http://pastebin.com/)
- [ ] Post full version information

Please make sure you also read [contribution guide](https://github.com/RuudBurger/CouchPotatoServer/blob/develop/contributing.md#issues) and followed all the steps. 
Make the title describe your issue. Having "CP not working" or "I get this bug" for 100 issues, isn't really helpful. My master will close issues if there isn't enough information. On a good day he will tag the issue on close with the reason (like `can't reproduce`), but usually he won't, the lazy asshat.

Sometimes my master seems like a grumpy cat and responds with short answers. This isn't (always) because he hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let him know. Maybe he can teach me his awesome ways.

Thanks!

RuudBurger:
Could you provide some example release already out there for testing?

RuudBurger:
Can't find any sample movies to test this out. Please let me know when there's content

cold12:
@RuudBurger there is now some content available to test from.

Ex:
Spider-Man 2 2004 2160p WEB-DL FLAC 5 1 x264-TrollUHD
Looper 2012 2160p WEB-DL FLAC 5 1 x264-TrollUHD

pudds:
Another vote for this request here as well.  There's not much content now, but I'd like to be able to set this as my preferred quality so it's picked up as it exists and becomes more common.

cold12:
Per the guidelines I'm informing you all that I am working on this feature and will issue a pull request.

webbson:
I would also love this now. 

I've seen some releases for recent big movies available in 2160p.

RedsGT:
Another vote for this feature.

Confusingboat:
Another vote here.

If you need to create a torrent/whatever for testing, a completely free, open source movie to use would be Tears of Steel.

https://mango.blender.org/download/

http://www.imdb.com/title/tt2285752/

williestwonka:
[...]

PR #5887, opened 2016-01-06
Add 2160p quality support

cold12:
Resolves issue #4596 .

I added an app.load event to the QualityPlugin to fill the qualities every time the app loads. This is so that new qualities that are added after someone has run the application once will still show up. I tested this and didn't encounter any issues.


RuudBurger:
Thanks!

PR #5905, opened 2016-01-09
Update newznab.py to use the newznab nzb link

boranblok:
If the newznab provider provides an nzb link xml tag this is used instead of constructing a link.
If no link tag is found in the newznab response the old logic is used.

This is mainly useful if the newznab provider passes links to nzbs that are not in the same domain as the api itself. (A content delivery network for example)


RuudBurger:
Thanks!

PR #5935, opened 2016-01-14
Fixed rating.join is not a function issue

moptel:
Fixed rating.join is not a function issue/IMDB top 10 box office chart not showing


RuudBurger:
Thanks!

PR #5966, opened 2016-01-19
Restoring of the lost code

maxkoryukov:
During previous resolving of conflicts in my pull requests, a few code was lost. There was a check for writing `non-writable` options to config.

Here that code with appropriate tests.


maxkoryukov:
Wooow! Superfast merge)))) Thanks)

PR #6005, opened 2016-01-27
Add t411 provider

ibibah:
Add t411 provider :
- Using t411 api
- Trying to respect couchpotato class model.

### DIFF
--- a/couchpotato/core/media/_base/providers/torrent/t411.py
+++ b/couchpotato/core/media/_base/providers/torrent/t411.py
@@ -1,0 +1,282 @@
+from couchpotato.core.helpers.variable import tryInt
+from couchpotato.core.logger import CPLog
+from couchpotato.core.helpers.encoding import simplifyString, tryUrlencode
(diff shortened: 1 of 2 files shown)


RuudBurger:
I won't add French only provider, sorry.
You can add it to your custom_plugin folder (which you can find in your data directory) though.

duramato:
@RuudBurger T411 inst a french only provider i use it all the time to get scene releases...

RuudBurger:
The site is and looking by the categories a lot of mixed content, which will lead to lots of false positives.

hadim:
@RuudBurger don't you think that it could be disable by default ? We can also add a warning message saying that this provider mostly have french content.

hadim:
What a shame...

kwaio:
Agreed, that's a shame. provinding that kind of content-rich providers, despite being mainly non-english (although a lot of soft-subbed english items) would help a lot of users. that's the kind of reasons I don't use CP anymore =/

PR #6010, opened 2016-01-28
remove defunct imdb dvd rental chart

razorloves:
The IMDB dvd rental chart no longer exists.  They removed it from their site about a months ago.
The dvd rental link now just redirects to the box office link, resulting in couchpotato showing the
same movies in both charts.
https://getsatisfaction.com/imdb/topics/where-did-the-top-dvd-weekly-rental-charts-go


RuudBurger:
Thanks!

PR #6012, opened 2016-01-28
Awesome-HD change from .net to .me

darknessgp:
Awesome-HD has changed from awesome-hd.net to awesome-hd.me This change updates the url so that CP can correctly search AHD again.

awesome-hd.net currently redirects to awesome-hd.me as well, though it doesn't appear to be a redirect that CP can follow correctly.


RuudBurger:
Thanks!

PR #6022, opened 2016-01-30
workaround: network paths for guessit

SytheZN:
Guessit seems to get hung up with network paths.
Replacing the \\ with C:\ seems to do the trick

Not sure how this will affect other platforms though, testing needed.

Fixes #4128 (couldn't find the referenced report in the close message)


SytheZN:
Oops, redo on develop

PR #6023, opened 2016-01-30
	workaround: network paths for guessit

SytheZN:
Guessit seems to get hung up with network paths.
Replacing the \\\\ with C:\ seems to do the trick

Not sure how this will affect other platforms though, I'm running on Windows.

Fixes #4128 (couldn't find the referenced report in the close message)

### DIFF
--- a/couchpotato/core/plugins/scanner.py
+++ b/couchpotato/core/plugins/scanner.py
@@ -918,7 +918,7 @@ class Scanner(Plugin):
         guess = {}
         if file_name:
             try:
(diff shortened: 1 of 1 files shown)


RuudBurger:
This is a fix that won't work on most systems, and even on windows the mounts aren't always on c:

SytheZN:
Nay, it doesn't update the path itself.

It just changes the path passed to guessit so it doesn't hang when collecting the file info.

And it's running on my system right now: 
Screenshot

Issue #5976, opened 2016-01-20
qBittorrent Downloader - API usage out of date


some-guy-23:
Running into a few issues with qBittorrent. The first is that having authentication on the endpoint seems to break functionality. The only way I could get it working with CouchPotato was to use the qBittorrent option to "bypass authentication for localhost". Then I set a blank username/password in CouchPotato downloader settings.

Second issue is probably related, but the "test" functionality for qBittorrent seems to pass, regardless of what info I enter (for example random port). So, seems to be a bug there?

Last trouble spot is actually already listed here (API supports labels, but CouchPotato hasn't yet implemented them) - https://github.com/RuudBurger/CouchPotatoServer/issues/4857

CouchPotatoBot:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [ ] Fill in all the dots. They're there for a reason
- [ ] Post logs, either inline (for smaller logs) or using [Pastebin](http://pastebin.com/)
- [ ] Post full version information

Please make sure you also read [contribution guide](https://github.com/RuudBurger/CouchPotatoServer/blob/develop/contributing.md#issues) and followed all the steps. 
Make the title describe your issue. Having "CP not working" or "I get this bug" for 100 issues, isn't really helpful. My master will close issues if there isn't enough information. On a good day he will tag the issue on close with the reason (like `can't reproduce`), but usually he won't, the lazy asshat.

Sometimes my master seems like a grumpy cat and responds with short answers. This isn't (always) because he hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let him know. Maybe he can teach me his awesome ways.

Thanks!

mrNobody1337:
I think I have the same issue, however I can't find anything usable in my logs. The connection test works, but no torrents are added. I have posted the info requested below, but my main question is which **version of the qbittorrent api** is couchpotato currently using? Thanks in advance!

//I posted this in another topic and now adding it here. I cannot bypass localhost auth since the two apps are running on different machines.

# Steps to reproduce:

    Install qBittorrent v3.3.2 (current stable version)
    Grab torrent

# Information:
Movie(s) I have this with: all
Quality of the movies being searched: 720p, 1080p, best
Providers I use: Kickass, Piratebay, Rarbg
version: de141aae (17/11/2015, 22:32:28) (running on ubuntu server 14.04)
Qbittorrent 3.3.2 on windows 7

[...]

Issue #5018, opened 2015-06-02
Qbittorrent 3.2 introduced a new incompatible api


eddie4:
Hi,

Am running CP 27-5-2015 and qBittorrent v3.3.0alpha. In qbittorrent 3.2 (current stable) the api has been overhauled. Which means that CP is no longer able to upload torrents.

new api:
https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-Documentation
and
https://gist.github.com/pmzqla/7e5733dbecfc50ee4ecd/2ccf07b06a3bfb5c9c0ecf6a13b513b972e0c0bd#file-web-api-documentation

Current error:
>POST /command/upload HTTP/1.1
>Host: 172.16.120.204:8080
>Content-Length: 257293
>Accept-Encoding: gzip, deflate
>Accept: */*
>User-Agent: python-requests/2.6.0 CPython/2.7.10 Linux/4.0.0-1-amd64
>Connection: keep-alive
>Content-Type: multipart/form-data; boundary=7e87d50e2abb4635b6af00e459e7d0b6
>--7e87d50e2abb4635b6af00e459e7d0b6
>Content-Disposition: form-data; name="torrent"; filename="torrent"
>d8:announce36:udp://open.demonii........

Qbittorrent response:
<HTTP/1.1 403 Forbidden

It is likely because of the new authentication method in qbittorrent.

CP log says the following:

### Steps to reproduce:
1. Install qbittorrent 3.2 or above
2. Attempt to snatch a torrent

### Information:
Movie(s) I have this with: Multiple
Quality of the movie being searched: brrip
Providers I use: Not the issue torrent gets downloaded.
Version of CouchPotato: git:(RuudBurger:CouchPotatoServer master) befbf3cc (2015-05-27 15:48:29)
Running on: Ubuntu 14.04

### Logs:
```
06-02 16:47:40 DEBUG [.downloaders.qbittorrent_] Sending "xxxx" to qBittorrent. 
06-02 16:47:40  INFO [hpotato.core.plugins.base] Opening url: get http://torcache.net/torrent/xxxx.torrent, data: [] 
06-02 16:47:40 DEBUG [tato.core.plugins.release] Downloader result: {'status_support': True, 'id': '0a18229f50f6d1ebef5341211f2fc720exxxf', 'downloader': 'qBittorrent'} 
06-02 16:47:40  INFO [tato.core.plugins.release] Snatched "xxxx in brrip 
[...]

Issue #4972, opened 2015-05-19
qbittorrent can't download magnet links


Wootlessberger:
### Steps to reproduce:
1. Use qbittorrent as a downloader
2. Download a movie from a provider with magnet links.

### Information:
Movie(s) I have this with: Every movies
Quality of the movie being searched: Any quality
Providers I use: kat.ph, torrentz and TPB.
Version of CouchPotato: source:(RuudBurger:CouchPotatoServer master) c295b344 (2015-05-02 13:58:52)
Running on: Windows 2012 R2

I can manually add a magnet link from kat.ph to qbittorrent without any issues.

### Logs:
```
05-18 23:51:12 ERROR [ato.core._base.downloader] Failed converting magnet url to torrent: 666BD69D2BD5F944B66A71E7351C63384ABE3D13 
05-18 23:51:12  INFO [tato.core.plugins.release] Tried to download, but the "torrent_magnet" downloader gave an error 
```

CouchPotatoBot:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [x] Fill in all the dots. They're there for a reason
- [x] Post logs, either inline (for smaller logs) or using [Pastebin](http://pastebin.com/)
- [x] Please wrap inline logs inside ``` for readability
- [ ] Enabled debug logging in Settings > Advanced (restart CP, be sure to disable after the bug is fixed)
- [x] Post full version information

Please make sure you also read [contribution guide](https://github.com/RuudBurger/CouchPotatoServer/blob/develop/contributing.md#issues) and followed all the steps. 
Make the title describe your issue. Having "CP not working" or "I get this bug" for 100 issues, isn't really helpful. My master will close issues if there isn't enough information. On a good day he will tag the issue on close with the reason (like `can't reproduce`), but usually he won't, the lazy asshat.

Sometimes my master seems like a grumpy cat and responds with short answers. This isn't (always) because he hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let him know. Maybe he can teach me his awesome ways.

Thanks!

Wootlessberger:
I did "Pick one yourself" on x-men days of future past, tried 3 release and none worked.

With debug (might not have all 3 snatch...)
http://pastebin.com/NXykLCzS

Only the ERROR for an other movie (A most violent year) :
[...]

Issue #4857, opened 2015-04-18
qBittorrent downloader: add label?


emzero:
I don't know if qBittorrent API lets you define a label when adding a new torrent, but if it does, I certainly use that option to use both CP and SickRage.
Could you add it if possible? Trying to move from uTorrent to qBittorrent

CouchPotatoBot:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [ ] Fill in all the dots. They're there for a reason
- [ ] Post logs, either inline (for smaller logs) or using [Pastebin](http://pastebin.com/)
- [ ] Post full version information

Please make sure you also read [contribution guide](https://github.com/RuudBurger/CouchPotatoServer/blob/develop/contributing.md#issues) and followed all the steps. 
Make the title describe your issue. Having "CP not working" or "I get this bug" for 100 issues, isn't really helpful. My master will close issues if there isn't enough information. On a good day he will tag the issue on close with the reason (like `can't reproduce`), but usually he won't, the lazy asshat.

Sometimes my master seems like a grumpy cat and responds with short answers. This isn't (always) because he hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let him know. Maybe he can teach me his awesome ways.

Thanks!

Rouzax:
qbittorent does not support the adding of labels yet through the API.
See https://github.com/qbittorrent/qBittorrent/issues/44

Rouzax:
Setting the label through the API will probably be implemented in version 3.3
See: https://github.com/qbittorrent/qBittorrent/pull/3444

EraYaN:
This has now been implemented. CouchPotato can now add/update support.

some-guy-23:
Would love to see support for this in CouchPotato. Right now using BlackHole method instead of web API, however with that qBittorent doesn't allow you to set a label (which I'd like for organization + post processing).

emzero:
+1 
qBittorrent API added a method to set labels on torrents.
SickRage already implemented this.
Would love to see it in CP so I cant ditch uTorrent ;)

some-guy-23:
This is also working in Sonarr, just to add another example.

whizzoo:
If someone is willing to implement this, here's a pointer:

self._post('command/setLabel', data={'hashes': hash, 'label':label}, cookies = cookies)

achmetinternet:
[...]

PR #6035, opened 2016-01-31
qbittorrent api upgrade

sickness0666:
updated qbittorrent downloader to support new api, with label support, minimum qbittorrent version required is v3.2
closes: #5976, #5018, #4857, #4972


RuudBurger:
Thanks!

PR #6061, opened 2016-02-08
internally used symlinking method fixed

ferensz:
In the internally used symlink helper method the operation system's hardlinking method were called instead of the symlinking method call.


RuudBurger:
Thanks!

PR #6065, opened 2016-02-10
Don't fork when run as a systemd service

craigcabrey:
systemd is able to track services much better when it is the direct parent of a service process. Forking (via daemonize) hinders this ability.


RuudBurger:
Thanks!

PR #6072, opened 2016-02-11
Synology error handling

tonyMartin:
Improved the log details when using Synology DownloadStation as a downloader. I added messages for the API error codes when creating a task and changed some log levels to be in line with the type of message.

I'm not sure my platform is perfect for the full range of testing so maybe this needs a quick check, but the changes are supposed to be pretty basic, no structure, just printing.


jkaberg:
@tonyMartin Please rebase on develop branch

tonyMartin:
Ok I will do this in a cleaner way ;)

PR #6075, opened 2016-02-12
fix xbmc mpaa rating metadata

lsellens:
xbmc is unable to parse mpaa rating metadata unless it says "Rated"


RuudBurger:
Thanks

PR #6084, opened 2016-02-14
Send magnet links directly to rtorrent.

sdrycroft:
Send magnet links directly to rtorrent, rather than trying to download the torrent file and sending the data.

### DIFF
--- a/couchpotato/core/downloaders/rtorrent_.py
+++ b/couchpotato/core/downloaders/rtorrent_.py
@@ -156,9 +156,13 @@ class rTorrent(DownloaderBase):
 
         # Try download magnet torrents
         if data.get('protocol') == 'torrent_magnet':
-            # Send magnet URL to rTorrent
-            self.rt.load_torrent_simple(data.get('url'), 'url', start=True)
-            return True
+            magnet_url = data.get('url')
+            # Pull out the ID of the torrent
+            urn_btih_position = magnet_url.index('urn:btih:')
+            torrent_hash = magnet_url[urn_btih_position + 9:urn_btih_position+49]
+            # Send magnet URL directly to rTorrent
+            self.rt.load_torrent_simple(magnet_url, 'url', start=True)
+            return torrent_hash
 
         info = bdecode(filedata)["info"]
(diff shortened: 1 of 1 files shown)


sdrycroft:
Third time lucky, apologies for the previous two pull requests.

RuudBurger:
Thanks for the PR, but you let it return the wrong data like this. Also ignoring some options the user has set (pause download etc). Please check my comments

couchpotato/core/downloaders/rtorrent_.py
```diff
@@ -156,12 +156,13 @@ def download(self, data = None, media = None, filedata = None):
...
-                return False
-
-            data['protocol'] = 'torrent'
+            magnet_url = data.get('url')
+            # Pull out the ID of the torrent
+            urn_btih_position = magnet_url.index('urn:btih:')
+            torrent_hash = magnet_url[urn_btih_position + 9:urn_btih_position+49]
+            # Send magnet URL directly to rTorrent
+            self.rt.load_torrent_simple(magnet_url, 'url', start=True)
+            return torrent_hash
```
RuudBurger: It should return a download id of the downloader. like below `self.downloadReturnId(torrent_hash)`

couchpotato/core/downloaders/rtorrent_.py
```diff
@@ -156,12 +156,13 @@ def download(self, data = None, media = None, filedata = None):
...
-            if filedata is False:
-                return False
-
-            data['protocol'] = 'torrent'
+            magnet_url = data.get('url')
+            # Pull out the ID of the torrent
+            urn_btih_position = magnet_url.index('urn:btih:')
+            torrent_hash = magnet_url[urn_btih_position + 9:urn_btih_position+49]
+            # Send magnet URL directly to rTorrent
+            self.rt.load_torrent_simple(magnet_url, 'url', start=True)
```
RuudBurger: This should be below and in an if/else statement with the `torrent = self.rt.load_torrent(filedata, verify_retries=10)`

Returning the torrent request, check if it download succeeded and set label/directory/paused

Issue #5970, opened 2016-01-19
Address already in use


byackee:
### Steps to reproduce:
1. ..
2. ..

### Information:
Movie(s) I have this with: ...
Quality of the movie being searched: ...
Providers I use: ...
Version of CouchPotato: git:(RuudBurger:CouchPotatoServer master) de141aae (2015-11-17 22:32:28)
Running on: ...

### Logs:
```
01-19 16:42:13 CRITICAL [                 __main__] Traceback (most recent call last):
  File "/app/couchpotato/CouchPotato.py", line 133, in <module>
    l.run()
  File "/app/couchpotato/CouchPotato.py", line 89, in run
    runCouchPotato(self.options, base_path, sys.argv[1:], data_dir = self.data_dir, log_dir = self.log_dir, Env = Env)
  File "/app/couchpotato/couchpotato/runner.py", line 335, in runCouchPotato
    server.listen(config['port'], config['host'])
  File "/app/couchpotato/libs/tornado/tcpserver.py", line 125, in listen
    sockets = bind_sockets(port, address=address)
  File "/app/couchpotato/libs/tornado/netutil.py", line 147, in bind_sockets
    sock.listen(backlog)
  File "/usr/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
error: [Errno 98] Address already in use
 
```

CouchPotatoBot:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [ ] Fill in all the dots. They're there for a reason
- [ ] Post logs, either inline (for smaller logs) or using [Pastebin](http://pastebin.com/)
- [x] Post full version information

Please make sure you also read [contribution guide](https://github.com/RuudBurger/CouchPotatoServer/blob/develop/contributing.md#issues) and followed all the steps. 
Make the title describe your issue. Having "CP not working" or "I get this bug" for 100 issues, isn't really helpful. My master will close issues if there isn't enough information. On a good day he will tag the issue on close with the reason (like `can't reproduce`), but usually he won't, the lazy asshat.

Sometimes my master seems like a grumpy cat and responds with short answers. This isn't (always) because he hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let him know. Maybe he can teach me his awesome ways.

Thanks!

maxkoryukov:
[...]

PR #6089, opened 2016-02-15
Improved handling of "Address already in use".

maxkoryukov:
Additionally, the startup-try-counter is repaired

fix CouchPotato/CouchPotatoServer#5970


RuudBurger:
Thanks again for all the improvements, hope your service does it thing with it ;)

PR #6093, opened 2016-02-16
Update Downloaders and Searcher page / Update old/dead link

razorloves:
Mostly just making links open in new window or tab.
Added a couple descriptions that were missing on searchers and downloaders.
Updated some old and/or dead links.


RuudBurger:
Thanks!

PR #6111, opened 2016-02-21
Allow listen on domain socket

psychomario:
If 'host' in the config file begins with `unix:`, port is ignored and the UI listens on the specified path as a unix domain socket.

Testing can be done with:
```
socat TCP-LISTEN:1234,reuseaddr,fork UNIX-CLIENT:/tmp/test.sock 
```

The use case I have in mind is breaking out of network namespaces without creating tunnels or anything. The socket will be connected to via nginx or similar.

It might be an idea to disable launch_browser, as it won't work.

### DIFF
--- a/couchpotato/runner.py
+++ b/couchpotato/runner.py
@@ -23,6 +23,7 @@ import requests
 from requests.packages.urllib3 import disable_warnings
 from tornado.httpserver import HTTPServer
 from tornado.web import Application, StaticFileHandler, RedirectHandler
+from tornado.netutil import bind_unix_socket
 from couchpotato.core.softchroot import SoftChrootInitError
 
 def getOptions(args):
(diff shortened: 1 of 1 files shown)


RuudBurger:
But if you have nginx, why not just proxy it?

psychomario:
When Couchpotato is in a network namespace, ie with only a VPN connection, nginx on a local network won't be able to connect to couchpotato. With this nginx can talk to the application, allowing UI access, but all other application traffic would be restricted inside the namespace.

RuudBurger:
Ok, it doesn't hurt the normal startup, so should be fine then

JasonMeudt:
Looks like this commit breaks a git install on Win10...

Commenting out line 26 "from tornado.netutil import bind_unix_socket" allows the startup to proceed on Win systems...

psychomario:
@JasonMeudt can you give [this branch](https://github.com/psychomario/couchpotatoserver/tree/winfix) a try?

([diff](https://github.com/CouchPotato/CouchPotatoServer/compare/develop...psychomario:winfix?expand=1))

JasonMeudt:
Starts up... Looks good so far...

psychomario:
OK, I'll submit that branch as a PR when I get home.

JasonMeudt:
Restarted CP and had these errors:

2-23 12:37:41 ERROR
[ couchpotato.runner] Failed starting: Traceback (most recent call last): File "C:\CouchPotatoServer\couchpotato\runner.py", line 367, in runCouchPotato loop.start() File "C:\CouchPotatoServer\libs\tornado\ioloop.py", line 783, in start for callback in callbacks: TypeError: 'NoneType' object is not iterable
02-23 12:37:41 CRITICAL
[ __main__] Traceback (most recent call last): File "C:\CouchPotatoServer\couchpotato.py", line 135, in <module> l.run() File "C:\CouchPotatoServer\couchpotato.py", line 89, in run runCouchPotato(self.options, base_path, sys.argv[1:], data_dir = self.data_dir, log_dir = self.log_dir, Env = Env) File "C:\CouchPotatoServer\couchpotato\runner.py", line 367, in runCouchPotato loop.start() File "C:\CouchPotatoServer\libs\tornado\ioloop.py", line 783, in start for callback in callbacks: TypeError: 'NoneType' object is not iterable

psychomario:
Curious, are you using the default config? ie with no `host` line set?

PR #6151, opened 2016-03-03
Updated utorrent.py

ksya:

This is to fix that CouchPotato unwillingly sets options that are not needed and may even irritate the user and fuckup other torrents not managed by CouchPotato. These settings can be set by the user if needed (which in most cases is not).
Setting the "read only after torrent finished" option does not even seem to be altered this way. I still had it to True and never had any problems. 
### Related issues:

This fixes issue #4411

### DIFF
--- a/couchpotato/core/downloaders/utorrent.py
+++ b/couchpotato/core/downloaders/utorrent.py
@@ -78,20 +78,6 @@ class uTorrent(DownloaderBase):
         if not settings:
             return False
 
-        #Fix settings in case they are not set for CPS compatibility
-        new_settings = {}
-        if not (settings.get('seed_prio_limitul') == 0 and settings['seed_prio_limitul_flag']):
-            new_settings['seed_prio_limitul'] = 0
-            new_settings['seed_prio_limitul_flag'] = True
-            log.info('Updated uTorrent settings to set a torrent to complete after it the seeding requirements are met.')
-
-        if settings.get('bt.read_only_on_complete'):  #This doesn't work as this option seems to be not available through the api. Mitigated with removeReadOnly function
-            new_settings['bt.read_only_on_complete'] = False
-            log.info('Updated uTorrent settings to not set the files to read only after completing.')
-
-        if new_settings:
-            self.utorrent_api.set_settings(new_settings)
-
         torrent_params = {}
         if self.conf('label'):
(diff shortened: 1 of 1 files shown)


mano3m:
if these options are not set, seeding does not stop after the time/ratio set in settings, also utorrent sets the readonly flag by default messing up every "move"-option except for hard-lync. That in your case these settings are not required doesn't mean they are not required in other set-ups. If you really need this I would suggest to make it an option, but:
- if you set the default share ratio really high in utorrent other torrents just keep seeding
- why would you want the files to be read only?

ksya:
> if these options are not set, seeding does not stop after the time/ratio set in settings
Might be true, but if this is what the user wants (me), CouchPotato should not interfere with this. 

> also utorrent sets the readonly flag by default messing up every "move"-option except for hard-lync.
True. Though the current implementation does not change this setting in uTorrent (see the comment in the code). Also in my opinion, if the user wants to use the "move" option, he can refer to documentation/Google after he finds out that it does not work. 

> if you set the default share ratio really high in utorrent other torrents just keep seeding
This is my current workaround :)

> why would you want the files to be read only?
I don't want to 'damage' the files by accident when uTorrent is still seeding them. Also, read-only is the default setting.

Thanks for looking into this.

RuudBurger:
I agree that these settings should be made an option, instead of removing them completely. It's annoying that these are global options, but there isn;t another way to this this on a per-torrent basis I guess.
Can't merge this PR though, so please extend it with settings in the downloaders > settings.

ksya:
Okay I added the option, but I'm a git noob and I can't figure out how to add this: https://github.com/ksya/CouchPotatoServer/commit/f568556e0d408d66f0ccfa348732579f09f154e5

This is the commit to my fork. Can someone of you add this (when there are no problems with this)?

maxkoryukov:
> Okay I added the option, but I'm a git noob and I can't figure out how to add this: ksya@f568556

Git and GitHub will make it all for you. Look at your repo on github.com : now there is a branch, related to current PR - https://github.com/ksya/CouchPotatoServer/tree/patch-1 . All changes in this branch come to PR automatically

Here is a similar question on stackoverflow: http://stackoverflow.com/a/9790530/1115187

**In two words**: commit and push your changes to the https://github.com/ksya/CouchPotatoServer/tree/patch-1

PR #6183, opened 2016-03-14
Fix Transmission directory when not running local

jvandenbroek:
When using a remote Transmission server, the directory parameter isn't set correctly due to a failed path existing check. This will only perform the check when host is localhost or 127.0.0.1. Not sure how to determine if set IP is actually local, so please enhance this if possible.


jvandenbroek:
Sorry, noticed too late that it needs to be in develop branch. Will do another pull request.

PR #6184, opened 2016-03-14
Fix Transmission directory when not running local

jvandenbroek:

When using a remote Transmission host, the set directory check will fail. This fix will only check the directory if host is localhost or 127.0.0.1. This probably could be enhanced by some socket function to check if supplied host is local, but this seems to work fine.

### DIFF
--- a/couchpotato/core/downloaders/transmission.py
+++ b/couchpotato/core/downloaders/transmission.py
@@ -67,7 +67,8 @@ class Transmission(DownloaderBase):
         }
 
         if self.conf('directory'):
-            if os.path.isdir(self.conf('directory')):
+            host = cleanHost(self.conf('host')).rstrip('/').rsplit(':', 1)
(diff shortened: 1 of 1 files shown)


NargiT:
This should solve the following issue #6370

couchpotato/core/downloaders/transmission.py
```diff
@@ -67,7 +67,8 @@ def download(self, data = None, media = None, filedata = None):
         }
 
         if self.conf('directory'):
-            if os.path.isdir(self.conf('directory')):
+            host = cleanHost(self.conf('host')).rstrip('/').rsplit(':', 1)
```
NargiT: What about http and https ?

piotrasd:
+1

RuudBurger:
Sorry, must have overlooked this one. Seems simple enough

Issue #6249, opened 2016-03-29
unable to run CP on synology DSM 6


pamporo:
### Steps to reproduce:
1. Upgrade Synology  415+ to DSM 6.0-7321
2. Open package center and click run on CP
3. Get the error "Failed to run package service". CP remains stopped.

### Information:
Movie(s) I have this with: all
Quality of the movie being searched: all
Providers I use: IPT
Version of CouchPotato: 20140604-4
Running on: Synology DSM 6.0-7321

### Logs:
messages
http://pastebin.com/raw/VRLfvMD4

synopkg
http://pastebin.com/raw/ddG1sUYM

CouchPotatoBot:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [x] Fill in all the dots. They're there for a reason
- [x] Post logs, either inline (for smaller logs) or using [Pastebin](http://pastebin.com/)
- [ ] Post full version information

Please make sure you also read [contribution guide](https://github.com/RuudBurger/CouchPotatoServer/blob/develop/contributing.md#issues) and followed all the steps. 
Make the title describe your issue. Having "CP not working" or "I get this bug" for 100 issues, isn't really helpful. My master will close issues if there isn't enough information. On a good day he will tag the issue on close with the reason (like `can't reproduce`), but usually he won't, the lazy asshat.

Sometimes my master seems like a grumpy cat and responds with short answers. This isn't (always) because he hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let him know. Maybe he can teach me his awesome ways.

Thanks!

andreifilip:
Hi @pamporo 

Had the same issue, and what I did was starting the server manually using the CLI and root privileges
Please use this only if you understand and know the risks when using the CLI.

I have 20160224 CP version

Also if you don't want or don't know how to do this, disregard my comment.

` /[yourVolume]/@appstore/[yourPythonPath_here]/bin/python /[yourVolume]/\@appstore/[YourCouchPotatoServer]/var/CouchPotatoServer/CouchPotato.py --console_log --config_file=/[yourVolume]/\@appstore/[YourCouchPotatoServer]/var/settings.conf`

Hope this help

pamporo:
Hi @andreifilip 

thanks! I managed to launch CP with your script. I still see CP as stopped in the package center though, do you know why? Also, if I close putty after launching it CP stops as well so I added & disown at the end of the script.

[...]

PR #6273, opened 2016-04-07
Fix start-stop script for DSM6 and beyond

simenon:
Fixes start-stop script for Synology users with DSM version 6.0 or higher. Also made it backwards compatible for older versions.

### Related issues:
https://github.com/CouchPotato/CouchPotatoServer/issues/6249
https://github.com/SynoCommunity/spksrc/issues/2104


simenon:
Closing this, as the issue seems to be more in user creation, then using sudo.

PR #6280, opened 2016-04-09
Fix for Synology DSM 6 devices

Producermark:
This will fix the bug of Couch Potato not starting on Synology's DSM 6


RuudBurger:
Thanks!

PR #6470, opened 2016-06-05
Add the ability to set default runtime parameters via config file

rigrassm:

Adds Feature

### Related issues:

N/A

## Feature Description

Currently, If a user wishes to customize their runtime configurations
the only way to do so is by passing in arguments via commandline
parameters each time the service is started. Since these options are
unlikely to change for an individual user, a configuration file able to
be parsed at runtime to set these options can be added to the
CouchPotatoServer application root directory. Once configured, the user
no longer has to pass any arguments via the commandline.

## Config File Setup

An example template for users too customize can be found at
`CouchPotatoServer/init/runtime_config.ini`

An example configuration:
```
[default]
daemon:
data_dir: /home/couchpotato
config_file: /etc/couchpotato/settings.conf
pid_file: /home/couchpotato/cp.pid

```

This example is equivilant to executing:
```
python CouchPotato.py --daemon \
					  --data_dir /home/couchpotato \
					  --config_file /etc/couchpotato/settings.conf \
					  --pid_file /home/couchpotato/cp.pid
```

For boolean parameters such as --daemon or --debug, provide just the
name of the option on its own line with or without the ":" on the end.

Parameters that have corresponding arguments require the name of the
parameter, a ":", followed by a space and the argument to be passed in.

Lines beginning with a # are considered comments and not parsed.

## Note

Parameters passed as command line arguments WILL take precedence over
the one specified in the config.ini file.

## Changes

* added package argparse_config.py to the lib folder
* Modified runner.py to parse config.ini file
* Added config.ini file which sets daemon mode by default
* Added init/runtime-config.ini which contains all available options with comments


RuudBurger:
Why is runner.py a new file?

rigrassm:
I'll look at it, I added a couple lines to it but it shouldn't be showing
as a new file.

rigrassm:
Fixed the issue and created a new pull request. See #6531.

PR #6495, opened 2016-06-12
Always select the first custom category as default

hashworks:
This fixes #3142, a 2 year old issue of mine. At that time the description actually stated `First one will be default.`, but someone removed that.

This change brings some advantages over the current code (`None` as default category):
- Foreign users can manage multiple languages, f.e. one can force German language by default but can select other languages when needed
- One can disable AC3 releases by default, but drop that limit when needed

Theoretically one could always select their default category manually, but this doesn't work for the automations. Also it's a pain in the ass to do that every time you search.

The downsite is that all users that use categories now will have their first category selected as default.
However they can restore the old behaviour by adding an empty category as their first category.


hashworks:
I'm not 100% sure about the ` and (status != 'done' or params.get('ignore_previous', False))` part. I copied that from the default profile, could someone explain to me what this exactly is for and if it makes sense at that position?

hashworks:
I fucked up, closing.

Issue #6370, opened 2016-05-07
In case of remote server, Transmission directory should not be validated on local machine


NargiT:
Hello,

I'm using [Docker on Raspberry Pi](http://blog.hypriot.com/) for my set up, and I encounter this issue.

### Steps to reproduce:
1. Set up 2 servers : First one with Couchpotato (let's call it S1) and a second one withTransmission (let's call it S2)
2. Configure CouchPotato to use Transmission
3. Settings > Downloaders > Transmission > Directory > PATH_TO_DOWNLOAD
4. Configure Transmission to use a default folder for torrents : DEFAULT_PATH
5. With docker ability to share Volumes (or using samba for physical servers) Couchpotato will be able to process finished movies
5. BUT since PATH_TO_DOWNLOAD is not present on S1 I get the following error (see logs)
6. When a movie is snatched, Transmission fallback to the DEFAULT_PATH (which is not visible by CouchPotato)

### Information:
Movie(s) I have this with: All
Quality of the movie being searched: All
Providers I use: Any
Version of CouchPotato: git:(CouchPotato:CouchPotatoServer master) 13a19423 (2016-05-29 09:30:23)
Running on: Docker with (http://blog.hypriot.com/downloads/)

### Logs:
```
05-07 06:30:32 ERROR [.downloaders.transmission] Download directory from Transmission settings:
/var/lib/transmission-daemon/completed_movies/ doesn't exist 
```

### Others
BTW I have the same set up for Sickrage, and everything works. I think the real problem is that the directory specified, should not be checked if the torrent client is a remote server.

file : couchpotato/core/downloaders/transmission.py
```python
if self.conf('directory'):
  if os.path.isdir(self.conf('directory')):
    params['download-dir'] = self.conf('directory').rstrip(os.path.sep)
  else:
    log.error('Download directory from Transmission settings: %s doesn\'t exist', self.conf('directory'))
```

CouchPotatoBot:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [x] Fill in all the dots. They're there for a reason
- [x] Post logs, either inline (for smaller logs) or using [Pastebin](http://pastebin.com/)
- [x] Please wrap inline logs inside ``` for readability
- [ ] Enabled debug logging in Settings > Advanced (restart CP, be sure to disable after the bug is fixed)
- [ ] Post full version information

[...]

PR #6524, opened 2016-06-18
DEVELOP branch : Fixe #6370

NargiT:
For remote Transmission client, the directory path is not validated by Couchpotato before to send the request.

### Related issues:
#6370 

In case of remote server, Transmission directory should not be validated
before to send the torrent


NargiT:
May I know whay it was closed ?

PR #6531, opened 2016-06-20
 Add the ability to set default runtime parameters via config file

RickyGrassmuck:
## Feature Description

Currently, If a user wishes to customize their runtime configurations
the only way to do so is by passing in arguments via commandline
parameters each time the service is started. Since these options are
unlikely to change for an individual user, a configuration file able to
be parsed at runtime to set these options can be added to the
CouchPotatoServer application root directory. Once configured, the user
no longer has to pass any arguments via the commandline.
## Config File Setup

An example template for users too customize can be found at
`CouchPotatoServer/init/runtime_config.ini`

An example configuration:

```
[default]
daemon:
data_dir: /home/couchpotato
config_file: /etc/couchpotato/settings.conf
pid_file: /home/couchpotato/cp.pid

```

This example is equivilant to executing:

```
python CouchPotato.py --daemon \
                      --data_dir /home/couchpotato \
                      --config_file /etc/couchpotato/settings.conf \
                      --pid_file /home/couchpotato/cp.pid
```

For boolean parameters such as --daemon or --debug, provide just the
name of the option on its own line with or without the ":" on the end.

Parameters that have corresponding arguments require the name of the
parameter, a ":", followed by a space and the argument to be passed in.

Lines beginning with a # are considered comments and not parsed.
## Note

Parameters passed as command line arguments WILL take precedence over
the one specified in the config.ini file.
## Changes
- added package argparse_config.py to the lib folder
- Modified runner.py to parse config.ini file
- Added config.ini file which sets daemon mode by default
- Added init/runtime-config.ini which contains all available options with comments


RuudBurger:
Thanks for the PR and sorry for my late response.
How is this different then using one of the startup scripts, and why aren't you using those? Are you on windows?

rigrassm:
Sorry for the late reply, must have missed the notification.

This design separates the configuration from the startup process all together.

It allows you to startup CouchPotato by simply calling the CouchPotato.py script. Whether you use a systemd unit file, an init.d startup script or just want to start it manually via the command line, it gets started using the same command(`python /path/to/CouchPotato.py`).

This makes for cleaner start up scripts that don't need to be modified anytime a user wants to make a change as simple as changing the applications data or log directories. Without using the config.INI, I have to:

1) open my systemd service file as root
    `sudo vim /etc/systemd/system/couchpotato.service`
2) change the argument passed into CouchPotato.py
3) Perform a daemon-reload
    `sudo systemctl daemon-reload`
4) Restart the service 
    `sudo systemctl restart couchpotato.service`

That's 3 commands that need to be run using sudo just to change the applications data directory. Using the config.ini, its reduced to:

1) open and edit the config.ini file(as a regular user, no privilege escalation necessary)
2) restart the service
    `sudo systemctl restart couchpotato.service`

It's not something that's done often but it makes things simpler when you do decide to make any changes or even if you just want to test a different configuration(like enabling debug mode, just uncomment the option in the config file and restart).

PR #6558, opened 2016-06-26
Update for qBittorrent 3.3.5

agentxan:
This should fix labels being added in qBittorrent 3.3.5. They changed labels to categories resulting in a change to the api. This should maintain backward compatibility with labels in previous versions. I tested with 3.3.4 and 3.3.5 on Windows 10 x64. When switching between versions I noticed that CouchPotato
did not re-authenticate with qBittorrent without restarting CouchPotato which I tried to fixed as well.
...


RuudBurger:
Thanks!

PR #6601, opened 2016-07-11
Fix for HD4Free not returning 'error' when there is no error...

swemoney:
I'm not sure if the HD4Free API used to always return an 'error' element in it's search response, but that doesn't appear to be the case anymore. The current HD4Free provider is broken because it fails to find 'error' when it searches for the login_fail_msg. This fix simply checks if 'error' exists in data before checking for the login_fail_msg.

### Related issues:
I think this was a change to the HD4Free API so I can't see there being any related issues.


swemoney:
Submitting a new pull request based on the develp branch...

PR #6633, opened 2016-07-21
Added Trusted only torrents option to TPB.

Tusky:
Enhances The Pirate Bay torrent provider with a Trusted only advanced option. When selected, only torrents from trusted uploaders will be downloaded.

### Related issues:
#6446


Tusky:
Sorry about pushing to `master`, but just checked `develop` and it's the same, so hopefully it won't cause any issues. I can re-push this against `develop` if needed.

RhyLlew:
Great idea, though I patched this into my version and it failed all results as untrusted. You may want to change the "or"s to "and"s. After that looks like it is working as expected.

RuudBurger:
#6754 merged agains develop. thanks!

Issue #6636, opened 2016-07-22
alpharatio search failing when ":" in movie name


lemosm:
THis is on development branch, didn't test on master.

### Steps to reproduce:
1. Find a movie with colon in the name
2. Search for the movie with alpharatio as the only searcher
3. find the search URL in the logs
4. copy and paste to browser, no results
5. remove the "%3A" from the URL, click enter, results 

my temp fix in alpharatio.py is to remove those 3 characters, I'm sure there is a better way but for anyone interested:
line 29 from:
`url = self.urls['search'] % self.buildUrl(media, quality)`
to:
`url = (self.urls['search'] % self.buildUrl(media, quality)).replace("%3A", "")`

this successfully fixes the issue.

CouchPotatoBot:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [x] Fill in all the dots. They're there for a reason
- [ ] Post logs, either inline (for smaller logs) or using [Pastebin](http://pastebin.com/)
- [ ] Post full version information

Please make sure you also read [contribution guide](https://github.com/RuudBurger/CouchPotatoServer/blob/develop/contributing.md#issues) and followed all the steps. 
Make the title describe your issue. Having "CP not working" or "I get this bug" for 100 issues, isn't really helpful. My master will close issues if there isn't enough information. On a good day he will tag the issue on close with the reason (like `can't reproduce`), but usually he won't, the lazy asshat.

Sometimes my master seems like a grumpy cat and responds with short answers. This isn't (always) because he hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let him know. Maybe he can teach me his awesome ways.

Thanks!

PR #6639, opened 2016-07-22
Fix Alpha Ratio searching with comma

Tusky:
Fixes searching with Alpha Ratio tracker. `:` kills their search engine and makes it return nothing at all. This simple solution by @lemosm works around that issue.

I'm unsure on the global direction, might be better to catch it in the `buildUrl` method, since a `:` sign should not be searched anyway IMHO.

### Related issues:
#6636


RuudBurger:
This is fine for now, thanks!

PR #6642, opened 2016-07-23
update qbittorrent.py for native magnet support

kdurrance:
Existing qBittorrent downloader required magnets to be converted to files, and failed frequently due to torcache availability. Updated downloader to support native magnet downloaders as already supported in the qbittorrent library class.

Never coded in python before - borrowed from working uTorrent downloader for inspiration, feel free to flame me but it works for me locally so thought it best to share.

### Related issues:
unable to download files with qBittorrent downloader due to torcache being unavailable.


RuudBurger:
Looks fine to me, thanks!

PR #6644, opened 2016-07-23
BitHUmen provider

Tusky:
### Description of what this adds:
A New search provider BitHUmen, which is a Hungarian private tracker.

### Related issues:
No issues.


RuudBurger:
I won't add language specific providers sorry.
You can add the provider in your custom_plugins folder (which is in your data folder). CP loads it from there after a restart.

Thanks for the code though!

Tusky:
Hey, it's not language specific, it's a private tracker, I'm using it like
that now though and though I would share. :)

PR #6835, opened 2016-09-05
Develop

guaguasi:

Torrentz.eu is dead. Reworked for Torrents2.eu, a clone.

_Includes cloudflare-scrape_ I noticed that in issue #4153 that you mentioned this isn't a good solution because cloudflare-scrape will need to be updated as needed and you can't include javascript engines. I feel the former is a non issue because as per the author, Cloudflare changes very rarely. The latter issue I assume has to do with licensing? Does https://github.com/PiotrDabkowski/Js2Py have that legal issue?
### Related issues:
#6810

### DIFF
--- a/couchpotato/core/media/_base/providers/torrent/torrentz2.py
+++ b/couchpotato/core/media/_base/providers/torrent/torrentz2.py
@@ -15,24 +15,18 @@ log = CPLog(__name__)
 class Base(TorrentMagnetProvider, RSS):
 
     urls = {
-         'detail': 'https://torrentz.eu/%s',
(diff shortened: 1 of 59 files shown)


ward0:
Thanks!

ward0:
When can you implement this one pls?

guaguasi:
I don't believe this will be implemented as it uses 3rd party library cloudflare-scrape to bypass the ddos protection. @RuudBurger said in #4153 that using that library int a good solution, but I don't know enough about Cloudflare and the bypass or have the time right now to engineer a different workaround.

As stated in the referenced issue, you could try talking to the people behind torrentz2.eu for a real API.

PR #6870, opened 2016-09-13
Add UHDBits provider

lansinghd:

This UHDBits provider is based off of the PTP provider and uses the Gazelle JSON API for searches.  

Searches are based on the title, year, quality, and media, but the API doesn't return the full release name for each result so returning information about group name, 3D, etc isn't available for now.  

Otherwise it seems to work well - any and all feedback is appreciated!

### DIFF
--- a/couchpotato/core/media/_base/providers/torrent/uhdbits.py
+++ b/couchpotato/core/media/_base/providers/torrent/uhdbits.py
@@ -1,0 +1,259 @@
+import htmlentitydefs
+import json
+import re
+import time
+import traceback
+
+from couchpotato.core.helpers.encoding import tryUrlencode
+from couchpotato.core.helpers.variable import getTitle, tryInt, mergeDicts, getIdentifier
+from couchpotato.core.logger import CPLog
+from couchpotato.core.media._base.providers.torrent.base import TorrentProvider
+from dateutil.parser import parse
+import six
+
+
+log = CPLog(__name__)
+
+
+class Base(TorrentProvider):
+
+    urls = {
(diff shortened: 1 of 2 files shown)


couchpotato/core/media/_base/providers/torrent/uhdbits.py
```diff
@@ -0,0 +1,259 @@
...
+        'login': 'https://uhdbits.org/login.php',
+        'login_check': 'https://uhdbits.org/inbox.php',
+        'api': 'https://uhdbits.org/ajax.php',
+        'detail': 'https://uhdbits.org/torrents.php'
+    }
+
+    login_errors = 0
+    http_time_between_calls = 2
+
+    def _search(self, media, quality, results):
```
RuudBurger: Please use the searchOnTitle function here, like https://github.com/CouchPotato/CouchPotatoServer/blob/master/couchpotato/core/media/_base/providers/nzb/omgwtfnzbs.py#L28
That function get's called multiple times with different title possibilities and does the work for you.

RuudBurger:
Thanks! If you fix the _search to the _searchOnTitle, I can merge it.

lansinghd:
Ah, I completely missed that function - thanks for the notes!  Fixed and tested, it successfully loops and tests multiple titles now.

theCheek:
I just joined this tracker and its amazing, so would love for it to be on the master branch (my seedbox doesn't allow for the switch to develop). Any idea when this update will merge with master?

rabidfurball:
me too!  Please merge this in.

PR #6884, opened 2016-09-17
Insert the name of the program being updated in the notifier

unreal4u:
This is useful in the case several programs use the same channel
to notify for updates, it is now possible to know which program
was updated within a simple view instead of guessing

### Related issues:
None

### Other information according to guidelines:
PR has not been tested, as the change is trivial. 

It doesn't have any limitations.

PR is up-to-date with the develop branch.


unreal4u:
Now develop is selected instead of master.

Issue #5961, opened 2016-01-18
Couchpotato 3.0.1 with Hadouken 5.2.0


maikelschippers:
I configured couchpotato to use a fallback to torrents when nzb can't be found. It does find a torrent, it tries to add it to hadouken 5.2.0 but nothing is added. I did adjust the config files properly

Couch gives the following errors:

### Steps to reproduce:
1. ..
2. ..

### Information:
Movie(s) I have this with: ...
Quality of the movie being searched: ...
Providers I use: ...
Version of CouchPotato: desktop: 3.0.1
Running on: ...

### Logs:
```
01-18 21:30:23  INFO [tato.core.plugins.release] Snatched "The Hunger Games 2012 BluRay 720p DTS x264-MgB [ETRG]": The Hunger Games (2012) in 720p from ThePirateBay 
01-18 21:31:58  INFO [hpotato.core.plugins.base] Opening url: get http://S03:8080/api?output=json&apikey=xxx&mode=queue, data: [] 
01-18 21:31:58 ERROR [   couchpotato.core.event] Error in event "download.status", that wasn't caught: Traceback (most recent call last):
  File "C:\Users\sysadmin\AppData\Roaming\CouchPotato\application\CouchPotato-3.0.1.win32\couchpotato\core\event.py", line 15, in runHandler
  File "C:\Users\sysadmin\AppData\Roaming\CouchPotato\application\CouchPotato-3.0.1.win32\couchpotato\core\_base\downloader\main.py", line 84, in _getAllDownloadStatus
  File "C:\Users\sysadmin\AppData\Roaming\CouchPotato\application\CouchPotato-3.0.1.win32\couchpotato\core\downloaders\hadouken.py", line 148, in getAllDownloadStatus
  File "C:\Users\sysadmin\AppData\Roaming\CouchPotato\application\CouchPotato-3.0.1.win32\couchpotato\core\downloaders\hadouken.py", line 412, in get_by_hash_list
TypeError: invoke() takes exactly 3 arguments (2 given)
[...]

PR #6888, opened 2016-09-18
update hadouken downloader to work with v5.2.0 api changes

wilreichert:
This [hadouken commit](https://github.com/hadouken/hadouken/commit/221d0153d21dec66b4d1fbc76a31ab8d81110134)  on 13 July 2015 refactored their v5 JSONRPC API. Looks like there were modifications on both the public facing bits as well as the JSON objects. Not sure I understand the rational moving dicts to arrays but whatever. Hadouken 5.0.0 uses the old v5 API, 5.1.0 and 5.2.0 uses the new. I've modified the couch downloader to work with the new v5 API. In retrospect perhaps I shouldn't have flat out broken the v5.0.0 API but this was mostly selfishly motivated. Happy to incorporate any feedback.

Their API is documented (more or less) [here] (https://github.com/hadouken/hadouken/tree/develop/js/rpc) and [here] (https://github.com/hadouken/hadouken/tree/develop/docs/src/for-developers/jsonrpc-methods)

Fixes
https://github.com/CouchPotato/CouchPotatoServer/issues/5961

### DIFF
--- a/couchpotato/core/downloaders/hadouken.py
+++ b/couchpotato/core/downloaders/hadouken.py
@@ -371,29 +371,29 @@ class TorrentItemv5(TorrentItem):
         self.obj = obj
 
     def info_hash(self):
-        return self.obj['infoHash']
+        return self.obj[0]
 
     def save_path(self):
-        return self.obj['savePath']
+        return self.obj[26]
 
     def name(self):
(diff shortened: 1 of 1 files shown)


RuudBurger:
Thanks!

PR #6964, opened 2016-10-08
Modify force renamer from hours to minutes. This allows finer grained…

ngodber:
Currently if user wants to force renamer to run more frequent then n Hours (1,2 etc) this is not possible. Fractional hours are not allowed as variable is recorded as int. By modifying unit from hours to minutes users who wish to do so can happily specify to the minute the frequency of forced renamer.

I use CouchPotato to rename and move movies that I have personally selected and downloaded (the completed download is moved to the watch folder). Waiting upto 1 hour for couch potato to recognise that a new movie has appeared is sub-optimal.

### Related issues:
Updating existing installs. I am not sure how the conversion from hour to minutes would be managed for users simply upgrading inplace. I do imagine that this is an edge case, most users would either use CP to managed the full download stack (in which case this value likely remains default) or are like myself and would probably prefer a more frequent option for force renamer.

### DIFF
--- a/couchpotato/core/plugins/renamer.py
+++ b/couchpotato/core/plugins/renamer.py
@@ -71,7 +71,7 @@ class Renamer(Plugin):
 
         fireEvent('schedule.remove', 'renamer.check_snatched_forced')
         if self.isEnabled() and self.conf('force_every') > 0:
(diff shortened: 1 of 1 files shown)


RuudBurger:
Thanks for the PR, but I think it's better to add another option, otherwise like you existing installs would be screwed.
Maybe add a toggle for "hour/minutes" and based on that fill in the schedule event.

PR #6998, opened 2016-10-13
Check that chmod is allowed before trying

### LINKED ISSUE #6995: chmod on utorrent downloads fails when utorrent is running as a different user
### Steps to reproduce:
1. Have couchpotato running as user couchpotato and utorrent as another user
2. Download movies through utorrent
3. Look for an "OSError: [Errno 1] Operation not permitted" error in the logs.

### Information:
Movie(s) I have this with: Most (but weirdly not all) movies downloaded as described above
Quality of the movie being searched: any
Providers I use: ...
Version of CouchPotato: 4a0999d9 (6-9-2016 11:53:22)
Running on: Debian 8.6

/couchpotato/core/downloaders/utorrent.py line 240 tries a chmod on the downloaded files, even when they are not owned by the user running couchpotato.
If the couchpotato user gets write permissions through e.g. group membership, this will fail.
If the renamer is setup to copy the files (as opposed to moving them) there's no need for write permissions at all as far as I can see. Only the folder must be writable in this case to accommodate the .ignore files.

mvn23:

Fixes an issue where couchpotato tried to chmod downloaded files without checking for permissions first
### Related issues:
#6995

Fixes #6995

### DIFF
--- a/couchpotato/core/downloaders/utorrent.py
+++ b/couchpotato/core/downloaders/utorrent.py
@@ -235,7 +235,8 @@ class uTorrent(DownloaderBase):
     def removeReadOnly(self, files):
         #Removes all read-on ly flags in a for all files
         for filepath in files:
-            if os.path.isfile(filepath):
+            #On *nix, check that we are allowed to chmod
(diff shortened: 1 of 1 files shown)


ngodber:
Can cofirm that this would be great as couchpotato is throwing errors due to this bug.

PR #7000, opened 2016-10-13
Added thepiratebay.org to piratebay proxy list

sulifluffy:

Added thepiratebay.org to the existing piratebay proxy list


dequire:
Hi @RuudBurger ,

Is this PR correct and will be merged?
Because The Pirate Bay proxy list found in couchpotato/core/media/_base/providers/torrent/thepiratebay.py really needs an update.
Most of the sites on this list stopped being actual mirrors and started injecting ads into magnet links, which breaks TPB in CouchPotato.

If this PR is too big for quick merging, I can create another clean one.

PR #7097, opened 2016-11-06
issue with none ascii chars in title name. decode to utf-8 and encode…

thetownfool:
... I had issues with automation when I turned on blu-ray.com integration today, unfortunately nothing was added to the watch list. After a little debugging it turned out a few days ago "Mötley Crüe: The End - Live in Los Angeles" was released and it was killing the cache check.

this fix just removes the none assci chars from the cache key. 

### Related issues:
...

… to ascii strips (for some reason the simple encode didn't work?

### DIFF
--- a/couchpotato/core/media/movie/providers/automation/base.py
+++ b/couchpotato/core/media/movie/providers/automation/base.py
@@ -45,7 +45,8 @@ class Automation(AutomationBase):
 
     def search(self, name, year = None, imdb_only = False):
 
-        prop_name = 'automation.cached.%s.%s' % (name, year)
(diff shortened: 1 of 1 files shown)


Matttman:
This didn't fix the issue for me but changing this did.

cache_name = name.decode('utf-8').encode('ascii', 'ignore')
to
cache_name = name.encode('ascii', 'ignore')

thetownfool:
Strange. I don't have this setup for debugging so my tests of what worked were simply at the shell level directly on the motley crue string. When only using encode('ascii','ignore') it still threw the error, not sure why. But then I've not touched python in over 10 years so it's very possible I was missing something

PR #7122, opened 2016-11-11
Display eta's

geogolem:
This fix displayed more clear ETA information for the movies in the wanted list.

It also updates the titles on the homepage with a better description of what they list.

Also, bypasses couchpotato api for eta information since it is returning nothing or the wrong result. It uses omdbapi instead. Also is commented code for using tmdb but requires tmdb api key to be specified in the code.

### Related issues:
...

### DIFF
--- a/couchpotato/core/media/movie/_base/static/details.js
+++ b/couchpotato/core/media/movie/_base/static/details.js
@@ -36,14 +36,18 @@ var MovieDetails = new Class({
 			)
 		);
 
-		var eta_date = parent.getETA('%b %Y') ;
+		var eta_date = parent.getETA('%b %d, %Y') ;
+		var dvd_date = parent.getDVDRelease('%b %d, %Y') ;
+		var theater_date = parent.getTheaterRelease('%b %d, %Y') ;
 		self.addSection('description', new Element('div').adopt(
 			new Element('div', {
 				'text': parent.get('plot')
 			}),
 			new Element('div.meta', {
 				'html':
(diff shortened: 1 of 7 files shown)


geogolem:
I hope I managed to get it right. I dont think the problem was what I was trying to merge.. but I wanted to use a change in another fork that has another pull request. My code actually is unaffected and does not require that change. That change is just a fix to prevent an error message from popping up all the time. The specific fix I am referring to is here: https://github.com/jpiron/CouchPotatoServer/commit/0119377169f9d2adad9d728fc798702b8716baa7

that fix has nothing to do with my changes; however, i need that fix to avoid the log always outputting some exception that that fix changes. Hopefully that guys fix will get pulled into develop at some point too.

geogolem:
I found a bug in this code that has a problem when trying to display ETA information for old movies. I believe its an easy fix.. but i have created another branch in my fork for a subsequent minor feature involving displaying whether or not a movie is available in netflix and the option to skip auto searching and downloading of titles that are currently available in a specific region's netflix library. I am going to make the fix regarding displaying ETA's for old movies in that branch.. and then initiate a pull request for that branch instead of this one.. Im going to close this pull request and I'll soon initiate a pull request for that branch.. since that branch will contain the changes from this pull request plus a few others + the fix for eta's of old movies.

geogolem:
I would have left this pull request.. and tried to just make the change in this specific branch. but most of the changes are localized to the same few files as the other branch. so I think its better and easier to just go ahead and continue working on the more up to date branch and initiate a pull request for that..

PR #7126, opened 2016-11-14
Feature: show if movie available on netflix + options to disable searching and/or downloading movies currently available on netflix

geogolem:
This is a follow-up to a previous pull request I subsequently canceled. In addition to displaying ETA data as specified in that pull request this also displays netflix information. Options were also added to the searcher so that the user can specify how that netflix information effects the searching (and subsequently downloading)

The default is: netflix information has no impact (it is only displayed)

Disable Auto download of netflix titles - (Search Only) - titles that are known to be currently available in netflix are searched but will not be automatically downloaded  -- a release will need to be manually downloaded via the dashboard

Disable searching of netflix titles - titles that are known to be currently available in netflix will simply be skipped during searches - downloads will not be possible with this option enabled.

-- there was an issue in that netflix information was not being updated when doing a global "search all wanted" -- this could lead to CP believing titles are in netflix when they are not, and subsequently skip searching/downloading them. I added an option "Always refresh before search" to ensure that when searches are performed ETA/netflix information is always up to date.

------------

Though a configuration option has not been implemented to allow the user to choose their netflix region. Code has been implemented to handle both Canada and US --> and others are possible in a similar fashion. the codes for each country need to be looked up on allflicks -- and the corresponding url must be updated to reflect the intended country. For now the code defaults to Canada, but a develop could switch to us simply by changing an assignment statement. A configuration option could be implemented to enable the user to specify the netflix region they want - but I couldn't fgure out how to get an option loaded from the config file in that part of the code. 

By default with all the changes in this pull request, it just displays more information than before and the behavior should remain unchanged. The code exists to modify the behavior based on the new netflix information, however, it  must be activated by the user via the configuration option. 

-------------

In a small minority of cases (less than 5%) netflix has different title's than IMDB. If this is the case CP will not identify a title as being available in netflix.

---> A good example of this is "Boychoir"  IMDB calls the movie "The Choir"
---> Another example is: The Death of Charlie Countryman

I personally didnt care that a few titles would not be marked as being on netflix when in fact they were. I never experienced a case where CP thought a movie was on netflix when it wasn't (I did during testing in a few cases, but I easily modified the code to make CP's criteria more strict). This could be improved perhaps with some kind of comma separated list in the configuration menu - but I dont think its necessary.

### Related issues:

Please refer to the previous (but now cancelled) pull request here: https://github.com/CouchPotato/CouchPotatoServer/pull/7122#issuecomment-259907055

--I think these changes would benefit from some UI improvements/additions - but only to make things prettier/better - they are fine the way they are. for example, it would be nice to have a tick box somewhere to hide all titles known to be available in netflix from view. Also a tickbox to enable/disable displaying netflix information or ETA information.

--Also, the feature would benefit by having a configuration option for the user to specify their netflix region. The code is written to handle this - but I just couldnt figure out how to get a configuration option loaded into the part of the code where it matters.

If someone wanted to help me implement the two points above - it would be greatly appreciated. Also I was unsure of how to make use of the existing "plugin" framework. Perhaps all this eta/netflix feature can be refactored to make use of that framework but I couldnt seem to figure out how to do it in that part of the code. Feedback and assistance is appreciated. The implemented code is fully functional - but it could benefit from improvement.


sneakyferris:
This sounds like a fantastic addition to me, great work dude!!!

ward0:
@geogolem  you should be allowed to work on CP! :) @RuudBurger

RuudBurger:
Thanks for the code and great addition. Need some time to get this merged and have some of the code in the proper place. Sorry for the slowness

ward0:
@RuudBurger  you should allow him to work on your project and help you out..

geogolem:
Allflicks xmlHTTPRequest queries started returning empty (which resulted in an invalid JSON error) unless a cookie was set with an identifier.

by viewing the page source of allflicks.net I found:
 j(document).ready(function() {
    var date = new Date();
    date.setTime(date.getTime()+(24*60*60*1000));
    var expires = "; expires="+date.toGMTString();
    document.cookie = "identifier=586e9d1655645"+expires+"; path=/; domain=.allflicks.net";


a one line change adding a cookie with that identifier to the session fixed the problem... hopefully this holds up over time. Looks like the expiry is set to 1000 days? not sure what this is .. but it resolved the issue with empty responses and invalid JSON errors.

geogolem:
So a few days later, the code started failing because the cookieID changed... not sure if I fixed in the best way, but now it dynamically grabs the cookieID from the site everytime.. This fixes the issue - but Im not sure if its the best way...

couchpotato/core/media/movie/providers/info/couchpotatoapi.py
```diff
@@ -102,12 +106,141 @@ def getInfo(self, identifier = None, adding = False, **kwargs):
...
+        if dvd_date != 'N/A' and year >1972:
+            p='%d %b %Y'
+            ddate=int(time.mktime(time.strptime(dvd_date,p)))
+           
+        if (ddate !=0):    
+            if ddate < tdate:
+                ddate = 0 #if the dvd release date occurs before the theater release date, assume the data is wrong
+                tdate = 0
+
+        if (ddate or tdate):
```
RhyLlew: Can i suggest making this an "and", so that the DVD release date will be grabbed if it is present in THEMOVIEDB but not OMDBAPI

geogolem: I see what you mean ---> changing to 'and' will mean it will accept OMDBAPI release date only if both theater and dvd dates are found... 

I think the reason I put "OR" was because in practice I found that if one was missing on one service, the probability of it being incorrect on the other service was high... and when I was writing the code I preferred no date over and incorrect date. 

Also, sometimes OMDBAPI would have partial information lacking in THEMOVIEDB.. 

I mean code can be written to cache the data and choose the most appropriate result in the end. I just didnt bother.. I will change it to "and" and try it out. and maybe write some improvements in the next few weeks.. so that we dont throw away perfectly good data from OMDBAPI because THEMOVIEDB is lacking.. or vice versa.. etc.

Thinking about it now and having been using this feature for a while, I suppose what you are suggesting makes more sense.. If a date is incorrect then someone could report it to the offending service (I did this a couple times and they did update their db).. I'll commit the change later this evening..

Please let me know if you have any further suggestions.. The Netflix feature was working flawlessly for the past several months until a few days ago when it seemed allflicks implemented some kind of cookie support. I checked in the update to "follow" allflicks change .. my ability to check netflix is totally dependent on allflicks....

RhyLlew: I only noticed this because I have found a release where TMDb DVD release date is present, but the OMDBAPI was not. Personally I only really care about the DVD ETA, so I thought it would be good that both services are checked. The way it was TMDb would probably never be checked for a DVD release date.

In all honesty I was only concerned with your great ETA gathering/display changes and don't have a Netflix account so I didn't pick up any issues there.

Thanks for your work by the way, it's good to be able to see in one place what stages each movie is at.

RhyLlew: You could also use the theatrical date and check if it is in the past to see whether you should bother checking the other service for a DVD release date.

geogolem: ok, i implemented some changes and briefly tested... could you try it out and see if it resolves your issue.. 

The way I implemented it now reflects the following:
        # if OMDBAPI returns both ddate and tdate use that and stop
        # if OMDBAPI returns partial or no information then check THEMOVIEDB
        # if THEMOVBIEDB returns both ddate and tdate use that and stop
        # if THEMOVIEDB returns partial or no information but OMDBAPI returns partial info use OMDBAPI's partial info and stop
        # if OMDBAPI returns no information and THEMOVIEDB returns partial information use that and stop
        # if both OMDBAPI and THEMOVIEDB return no information, no information is propagated and we stop.
        # could add some code in here to merge the two.. i.e. if one has dvd info and the other has theater info...
        # but ignore that case for now.

OMDBAPI takes precendence.. but full information takes precedence over partial information and there is no constructive merging of info from both sources... (i.e. to make full information out of two pieces of partial information)...

your suggestion regarding immediately accepting partial information if the theatrical release date is in the future is a good one.. (and i was about to implement it; however, i think its better to just check anyway, since if the theatrical release date is incorrect, and THEMOVIEDB has full information, it would be better) -- plus im too lazy to go in and add that check right now.. might do it in future if i have the time.... 

Please let me know if you have any further suggestions.

geogolem: hopefully your case doesn't fall into that "merging" info situation..

this would be an issue if OMDBAPI had only theater release date...

THEMOVIEDB had only dvd release date..

in this case, THEMOVIEDB would be checked, but the set of data returned from OMDBAPI would be used... (i.e. the theater release date would be used and the dvd release date would be discarded - its not that theater takes precedence over dvd, but that OMDBAPI takes precedence over THEMOVIEDB -- as it is implemented now).. 

it would be quite easy to add some kind of merging situation.. if one service has a dvd_date and the other has a theater_date.. then, make sure they are sane (i.e. theater_date < ddvd_date) and use them both instead of thowing one away..  I'll try that out righ tnow and commit - so dont bother testing the changes just yet.

geogolem:
@RhyLlew : regarding the discussion we were having... I originally commented there, but am moving it here: 

regarding skipping the check if the theatrical date is in the future --> i wont bother with that since, it is possible that a theatrical date is in the future but the dvd release date is known (so we still want to know it) and also, it could be possible that OMDBAPI is just wrong... while THEMOVIEDB might have more accurate or complete information.

Its a good suggestion though -- but, I think its not worth it..

Anyway, I have implemented the changes.. and am just testing them now.. Once I finish testing I will commit and you can try them out. (I checked in partial changes already, but then I realized I wanted to improve it further)....

As of the time of this comment (but not with the current committed code) it should behave this way:

    # if OMDBAPI returns both ddate and tdate use that and stop
    # otherwise OMDBAPI returned partial or no information so check THEMOVIEDB
    # if THEMOVBIEDB returns both ddate and tdate use that and stop
    # at this pointif we are still going, both OMDBAPI and THEMOVIEDB each returned no or partial information
    # if both OMDBAPI and THEMOVIEDB returned partial information and the partial info doesnt overlap construct full information
    # if the full information constructed makes sense, use that and stop
    # at this point we know full information is not going to be available
    # if OMDBAPI had partial information, use it and stop
    # if THEMOVIEDB returns partial information use it and stop
    # if we got here both OMDBAPI and THEMOVIEDB return no information, no information is propagated and we stop.

Please let me know if you run into any problems, or have any further suggestions!

geogolem:
I added a CouchPotatoTray.pyw launcher for Windows. which launches a tray icon alongside CouchPotato.py.. This is only for windows and it doesnt have any functionality... but its nice to be able to tell if CP is running via the tray icon.. if you dont want the tray icon just run like you normally would. If you want the tray icon launch using this pyw file.  dont really care if this minor change gets integrated.. its just that if you are on windows and running from source you lose the tray icon. Maybe I will add some features to make double click open CP dashboard etc. but for now, I just like having the tray icon there to show that the program is running. This may be buggy. I just took the py32 tray icon code example and made it work with CP.

geogolem:
so every once in a while it seems that allflicks updates their website and changes something minor that causes the netflix check in CP to break. Everytime this happens I chase the change down and update the code. I try to generalize the code so that future changes will not cause allflicks to break but rather the code will continue to function regardless of the change. I am doing the best I can -- but I intend to continue supporting/updating the feature as required (since I use it regularly)...

bitzorro:
How would this work related to geographic releaces? Otherwise i like the idea and it would be nice if it could send notifikations that the movie is actual on Netflix.

geogolem:
@bitzorro : 

Right now by default only netflix Canada is checked; however with a single line change in the code, it can be switched to US.

Other regions have not been implemented but would follow a similar fashion. there are very minor differences between checking US vs. Canada i.e. use allflicks.net/canada instead of allflicks.net .... though it has been implemented for US and Canada it would not be difficult to implement other regions as well.

I would eventually like to add a section in the Cp configuration where the user could specify which netflix region they are interested in; however it has not been done yet.

It would also be possible (although not implemented) to check multiple netflix regions and display the result as well.

geogolem:
This code all died, because allflicks changed to reelgood... 

reelgood only supports the US right now.. I might eventually tyr and update this, but for now this code is sort of useless..

PR #7128, opened 2016-11-15
Vuze Remote Support - Get X-Transmission-Session-Id from headers

kmicki:
**Reason**

This modification allows using Vuze Remote (Azureus) by configuring it as Transmission downloader in CouchPotato.

Vuze Remote json-rpc protocol is compatible with the one of Transmission, but there is a small difference with the HTTP Error 409 message that it returns when CouchPotato uses wrong (or none-initially) session ID. Transmission returns proper session ID inside the error message and the headers. Vuze Remote provides the session ID only in the headers of returned packet. Before the modification, the error message was scanned for new ID. After the modification the new session ID is acquired from the headers, so both Transmission and Vuze Remote works.

**What was done**

Changed the method of refreshing X-Transmission-Session-Id, when the one provided for connection is mismatched and HTTP Error 409 occurs.

When HTTP Error occurs,  _urllib2.urlopen_ throws an exception, so we cannot use the return value to read headers. Apparently there is an undocumented _hdrs_ element of _HTTPError_ object that we can use to read headers. That is used in the modification.

### DIFF
--- a/couchpotato/core/downloaders/transmission.py
+++ b/couchpotato/core/downloaders/transmission.py
@@ -237,7 +237,7 @@ class TransmissionRPC(object):
                 msg = str(err.read())
                 try:
                     self.session_id = \
-                        re.search('X-Transmission-Session-Id:\s*(\w+)', msg).group(1)
+                        err.hdrs['X-Transmission-Session-Id']
                     log.debug('X-Transmission-Session-Id: %s', self.session_id)
 
                     # #resend request with the updated header


stephen304:
I can't get this to work - I have vuze web remote installed and visible on :9091, and I get a json response from /vuze/rpc/json="". What do I put for rpc url in couchpotato? It keeps telling me 404 or incorrect password in the logs (I tried not using user/pass too)

kmicki:
@stephen304 

I set rpc url to:
`transmission`
It works for me. Of course you need to apply modifications from this PR to prevent error 409.

stephen304:
@kmicki Thanks! I managed to make it work, though I still don't exactly understand why it's working. Turns out Vuze responds on both /vuze/rpc and /transmission/rpc. The only issue I have now is that enabling password and entering it into CP causes the connection to fail due to incorrect password. But since I have other security measures, this should be fine for me.

I have one more question if you don't mind - how should I configure CP with regard to torrent files? I set Vuze to move torrent data on completion, but not the .torrent files. I then configured CP renamer to rename from the folder that Vuze moves torrent data to on completion, but it also has "Torrent file action". Do you know what I am supposed to do with this option?

PR #7133, opened 2016-11-23
Putio -- added ability to have https callback url and pulled new version of put.io api

dumaresq:
Added ability to set callback url for putio to https.  Also pulled in a new version of the putio api which allow for better downloading from putio site.


RuudBurger:
Thanks!

PR #7136, opened 2016-11-26
Add year to search param for torrentpotato

sweco-secfha:
Jackett returns only 50 results for each search, without searching on year it is impossible to get a good hit.

I'm not used to python and this was a quick fix, so i'm not sure if this is how you do it in python.

### DIFF
--- a/couchpotato/core/media/movie/providers/torrent/torrentpotato.py
+++ b/couchpotato/core/media/movie/providers/torrent/torrentpotato.py
@@ -17,6 +17,6 @@ class TorrentPotato(MovieProvider, Base):
             'user': host['name'],
             'passkey': host['pass_key'],
             'imdbid': getIdentifier(media),
-            'search' : getTitle(media),
+            'search' : getTitle(media) + ' ' + str(media['info']['year']),
         })
         return '%s?%s' % (host['host'], arguments)


maxcore:
very good point !!!!

seanvree:
YES PLEASE!

seanvree:
So I mad this change to the TorrentPotato.py file and seemed to work!!!

Can we make the same change to the Newznab file?

RuudBurger:
Isn't it default searching on IMDB id?

clinton-hall:
In the forums there are numerous reports of year been ignored. Not happening with my providers, but certainly Jacket and torrent potato seem to do this.

clinton-hall:
https://couchpota.to/forum/viewtopic.php?f=5&t=50127

seanvree:
@RuudBurger ... Nope, if the movie is not immididatly found by CP (aka, not released or "rar") CP has been snatching ANYTHING that is even close to the the name ignoring all other filters. I thought it was designed to simply not snatch it if it couldn't find the proper title, but if you read the topic linked above by @clinton-hall (which I started) you'll see that this isn't the case.  I have all the top tier private trackers on both Usenet side and Torrent side.  And in my case, I mostly got "false-positive" snatches with NZBs, NOT torrents, although I did implement this fix as a testing measure and sure enough it removed all the Torrent false positives.

Thanks for your work.

seanvree:
Also @RuudBurger , I have a very powerful windows-based server that has jackett, nzbhydra and all the top tier trackers set up. If you'd like to RDC into my machine via teamviewer or RDC for testing, I can facilitate that.

sweco-secfha:
@RuudBurger I only use iPtorrents and they don't seem to have any support for IMDB search.
@seanvree Great to get some support for this, hope it makes it into master soon.
I don't use Newznab or any nzb for that mater, but it seems by looking at the code that Newznab only ever searches by IMDB id so it's weird that you get false-positives there.

seanvree:
@sweco-secfha , I thought that too, but it's searching via name....and even doing a horrible job at that.

Nyghto:
I applied the change, and year is getting passed now into Jackett, which will already saves a lot of fakes!! but still some fakes can pass, example: Rings 2016, still gets results back like Kubo and to two String 2016.

seanvree:
@Nyghto , right, but WAAAAY better than before! Just need to get this changed for Newsnab providers!

PR #7140, opened 2016-12-06
Fix Transmission not reading correct status

fernandog:
Seeding status is 6. Not 16


fernandog:
Seed is not complete right? Someone change to 16 on purpose?

RuudBurger:
Shouldn't the whole line be remove then? Because it isn't "completed" when it's seeding.

fernandog:
I did some tests and also need to check isFinished because it set this when seeded completed. If seeded is not compleded but torrent is completed and paused it would wrong remove it.

fernandog:
removed that line and added the `isFinished` check

fernandog:
@RuudBurger not interested?

fernandog:
@RuudBurger can you please at least reply something?

PR #7150, opened 2016-12-22
add Accept: * header to urlopen

scambra:
Needed for magnetdl, it returns nothing if none Accept header is present, so magnetdl provider is not working, you can test with wget:

```
wget -O - http://www.magnetdl.com/t/tempting-fate-1998/se/desc/1/
wget --header="Accept: *" -O - http://www.magnetdl.com/t/tempting-fate-1998/se/desc/1/
```

Accept: * shouldn't break other providers.

### Related issues:
Same pull request #7149 against develop branch now


johnygreg:
Wouldn't it be better to just add the missing header directly in magnetdl.py?

Like this:
data = self.getHTMLData(url, headers = {'Accept': '*'})

That way we don't have to assume and it is sure to work with every other providers.

PR #7159, opened 2017-01-04
Add 2160p support to PTP provider

fenduru:
Fixes error in PTP provider when searching 2160p and allows search to complete successfully


RuudBurger:
Thanks!

PR #7161, opened 2017-01-09
Additional Kickass Torrent Proxies added

shawnmstout:


RuudBurger:
Thanks. Added manually without the IDEA file

PR #7164, opened 2017-01-10
Add option to configure search categories for Newznab providers

jwvanderbeck:

The Newznab API function that CouchPotato uses for searching for movies doesn't always search every category on some sites.  As a result some possible results will not be returned.  I personally ran into this using dognzb where I discovered that category 2060 Movies/3D was not included in the search.

This PR adds the ability for the user to configure custom categories lists to use in searches on Newznab providers.  

Beyond fixing this issue, it allows users to better target searches by eliminating categories they might never want to search in such as Movies/Foreign or Movies/SD for example.

This PR will add a new option to the NewzNab settings in Settings->Searcher called "Custom Category" that takes a **space** deliminated list of categories numbers to use in searches.  If set, the query URL will be appended with "cat=" and the list of categories in proper Newznab format.

### Related issues:

https://couchpota.to/forum/viewtopic.php?f=5&t=51270

Geesh everything I typed out vanished.. arg.  Give me a few and I will re-enter it.

### DIFF
--- a/couchpotato/core/media/_base/providers/nzb/newznab.py
+++ b/couchpotato/core/media/_base/providers/nzb/newznab.py
@@ -128,6 +128,7 @@ class Base(NZBProvider, RSS):
         api_keys = splitString(self.conf('api_key'), clean = False)
         extra_score = splitString(self.conf('extra_score'), clean = False)
         custom_tags = splitString(self.conf('custom_tag'), clean = False)
+        custom_categories = splitString(self.conf('custom_categories'), clean = False)
 
         list = []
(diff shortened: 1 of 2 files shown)


Fossil01:
'default': '2000,2000,2000,2000,2000,2000',

Maybe add some other default Newznab categories there? Like 2040 for HD, 2045 which is UHD, or 2080 for WEB-DL.

jwvanderbeck:
2000 is the top level so by specifying that, you search in ALL movie categories.  Basically that default is what you would expect CP to be doing now but isn't.

PR #7158, opened 2017-01-03
IPTorrents query fixes - From PR #7023

kmr2319:
Pulled just the IPT related fixed from PR #7023 and submitted this
against the develop branch.
* Improve IPTorrents quality to category mapping
* Remove obsolete query parameter for categories
* Remove quotes around movie title
* Search in title, not title and tags

In addition, changed the IPT domain to *.com - *.eu didn't seem to work at the time. *.eu *is* loading manually though.

### Description of what this fixes:
IPTorrents search results aren't very good without these changes. Sometimes not finding any movies at all - even though manual searches on IPT return the expected results.

I've tested this myself and have been running it for a week or so with great results. 

Provided changes to one other user as well and they had initial good results as well.

### Related issues:
PR #7023 

Hat tip to @tsoernes for the original PR

PR #7167, opened 2017-01-17
Removed unnecessary comma

kmr2319:
Did not cause any problems in my testing, but still doesn't need to be
there...

Unnecessary comma on line 19 as noted in #7158

### Related issues:
PR #7158


danielsamuels:
It's better to be there so that re-ordering lines doesn't cause unnecessary noise in future diffs.

RuudBurger:
What he said ;)
Thanks for the PR though

PR #7172, opened 2017-01-28
Bring back YTS/YIFY torrent provider.

emphaticsunshine:
Adding YTS/YIFY provider back. I have tested it out and it works great.


ward0:
@emphaticsunshine  can you share your .py files pls?

ward0:
Getting an error in : /static/scripts/combined.base.min.js

class is not defined: 

var Uniform = new Class({
    Implements: [ Options ],
    options: {
        focusedClass: "focused",
        holderClass: "ctrlHolder"
    },

Looking in static/scripts/combined.plugins.min.js

RuudBurger:
I've merged another YTS provider.
Thanks for the PR though!

PR #7173, opened 2017-01-29
Add configuration setting to make CP use the first category as default

hashworks:
This adds a new config setting `categories.first_as_default` (default `false`) which makes CouchPotato select the first category by default in search and automation. 

This change brings some advantages over the current code (None as default category):

- Foreign users can manage multiple languages, f.e. one can force German language by default but can select other languages when needed
- One can disable specific releases by default, but drop that limit when needed

Theoretically one could always select their default category manually, but this doesn't work for the automations. Also it's a pain in the ass to do that every time you search.

I'm not really happy with my `couchpotato/core/plugins/category/static/category.js` change, it seems like there is some automatism included which generates the settings tabs from the configuration settings in the `__init__.py` files. But I can't seem to get that running. Any ideas?


phillipjf:
Any updates on this? Looking for this functionality..

hashworks:
I switched to Radarr. The code in this PR should work but don't expect support from my side. It has been 3 years.

phillipjf:
@hashworks Thanks for the update. I'm setting up Ombi but need movies to get a default category applied. Does this work with Radarr?

hashworks:
Yes. Radarr has vastly more features.

PR #7175, opened 2017-02-01
Fixes case where sometimes getQuality returns null when empty object …

ktiedt:
If no quality matches, getQuality's filter call returns no matches, getQuality returns null when it should return an empty {}


ktiedt:
Just saw comment from the Bot saying to use Develop branch on another PR -- but the "Contributing Guide" link was broken so, no idea if there were any other concerns I should address.

ktiedt:
Ugh, closing this and opening a proper one w/o all the noise from having used the wrong base.

PR #7193, opened 2017-04-02
Fix OmdbApi parsing

TheQwertiest:
Fixed OmdbApi parsing: omdapi json also contains list (under 'Ratings' key), trying to call lower() on this list resulted in error.

### DIFF
--- a/couchpotato/core/media/movie/providers/info/omdbapi.py
+++ b/couchpotato/core/media/movie/providers/info/omdbapi.py
@@ -88,7 +88,8 @@ class OMDBAPI(MovieProvider):
 
             tmp_movie = movie.copy()
             for key in tmp_movie:
-                if tmp_movie.get(key).lower() == 'n/a':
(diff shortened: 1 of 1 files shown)


AdamBarnwell:
I can confirm this is as a current live issue which means you generate an error every time you search for a movie by typing in the search box as it doesn't know what to do with some of the resulting response from the OmdbApi.

Please merge ASAP so my logs aren't littered with errors.

Morphy2515:
Have the same error: 
[ie.providers.info.omdbapi] Failed parsing IMDB API json: Traceback (most recent call last): File "/home/morphy/.couchpotato/couchpotato/core/media/movie/providers/info/omdbapi.py", line 91, in parseMovie if tmp_movie.get(key).lower() == 'n/a': AttributeError: 'list' object has no attribute 'lower'

animosity22:
Same error here as well. Please merge :)

RuudBurger:
Thanks!

PR #7199, opened 2017-04-19
Report Issue Link 404s

couchuser12345:
Obviously this shouldn't be merged, but I'm not sure how else to bring attention to this since I can't file an issue.

Access the issue URL results in a 404 since issues appears to be not public or blocked?

If this is the case then filing issues from within the CouchPotato UI should probably be revisted.


couchuser12345:
cc @RuudBurger @YourHuckleberry 

Was issue tracking disabled on GitHub? Wanted to report a separate issue from the CouchPotato UI, but it leads to the above 404

RuudBurger:
Yeah, this is like intended. Because I don't have time to fix the 1000+ issues that are on there ;)
Sorry.

PR #7203, opened 2017-05-04
Fix Letterboxd integration and add multi-page support

pindab0ter:
Letterboxd integragion was broken, this is fixed in the first commit.

The second commit adds support for multiple pages. When a watchlist (like for example https://letterboxd.com/pindab0ter/watchlist/) has multiple pages, only the first page would be scraped.

I have not managed to figure out how to properly develop and test for CouchPotato, I also think I might have not adhered to the style.

Please test if this works and give feedback on what needs to be improved/changed before it can be merged.

### DIFF
--- a/couchpotato/core/media/movie/providers/automation/letterboxd.py
+++ b/couchpotato/core/media/movie/providers/automation/letterboxd.py
@@ -13,7 +13,7 @@ autoload = 'Letterboxd'
 
 class Letterboxd(Automation):
 
-    url = 'http://letterboxd.com/%s/watchlist/'
+    url = 'http://letterboxd.com/%s/watchlist/page/%d/'
     pattern = re.compile(r'(.*)\((\d*)\)')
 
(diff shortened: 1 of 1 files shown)


ntassone:
I have this running locally and my Letterboxd watchlist automation is fixed. 🎉

pindab0ter:
The accuracy of the matching could be increased. Currently only the name of the movie found in the `alt` tag of the poster image is used.

In order to get the year of release as well you would have to render the dynamic content using a web driver like Selenium before getting the source with BeautifulSoup. This would require adding Selenium as a new dependency, but would greatly increase the reliability of the findings.

@RuudBurger what are your thoughts on this?

pindab0ter:
@RuudBurger would like to hear your thoughts on this. As it stands this plugin generates a _lot_ of false positives because it only has the name (not the year) of the film to work with.

PR #7204, opened 2017-05-11
Updated script notification to use Popen instead of call.

Fmstrat:
My earlier PR that you merged used subprocess.call which I realized wasn't the same way you were calling other external functions in CouchPotato. For consistencies sake, I have converted it over to subprocess.Popen.

### Related issues:
n/a


RuudBurger:
Thanks!

PR #7206, opened 2017-05-20
Fixing error where casting .lower on non string

krisnova:

There was a bug where the code wasn't checking if an object was a string or not before casting `.lower()` on it.

### Logs:
```
05-20 05:56:43 ERROR [ders.torrent.torrentbytes] Failed to parsing TorrentBytes: Traceback (most recent call last):
  File "/opt/couchpotato/couchpotato/core/media/_base/providers/torrent/torrentbytes.py", line 56, in _searchOnTitle
    link = cells[1].find('a', attrs = {'class': 'index'})
IndexError: list index out of range
 
```

### DIFF
--- a/couchpotato/core/media/movie/providers/info/omdbapi.py
+++ b/couchpotato/core/media/movie/providers/info/omdbapi.py
@@ -88,8 +88,9 @@ class OMDBAPI(MovieProvider):
 
             tmp_movie = movie.copy()
             for key in tmp_movie:
-                if tmp_movie.get(key).lower() == 'n/a':
-                    del movie[key]
+                if type(tmp_movie.get(key)) is string:
+                    if tmp_movie.get(key).lower() == 'n/a':
+                        del movie[key]
 
             year = tryInt(movie.get('Year', ''))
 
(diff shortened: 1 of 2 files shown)


kris-nova:
Also just a heads up..  

The link to open up a new GitHub issue is broken (otherwise I would have done that)

```
https://github.com/CouchPotato/CouchPotatoServer/issues/new
```

.. Looks like somebody turned issues off for the repository.

couchpotato/core/media/movie/providers/info/omdbapi.py
```diff
@@ -88,8 +88,9 @@ def parseMovie(self, movie):
 
             tmp_movie = movie.copy()
             for key in tmp_movie:
-                if tmp_movie.get(key).lower() == 'n/a':
-                    del movie[key]
+                if type(tmp_movie.get(key)) is string:
```
ekohl: This is a bad practice. If anything you'd need to use ```isinstance``` with basestring on python 2 and just string on python 3 but it's likely you'll just want to check if it's true-ish or just catch a KeyError.

PR #7208, opened 2017-05-25
Use IMDb codes for search and comparison

jbarr13:
Using the movie title to search was inaccurate and often failed. Also, the original title is not always correct. YTS offers lookup by imdb code and always finds an exact match if they have it. This also improves speed.

Searching YTS by movie title sometimes returned no results. Sometimes the results returned by the search were marked as incorrect even though they were correct. The search now uses the movies IMDb id.

### Tests:
I have tested this by comparing the results of searching for various movies. A few of these were "X2", "I.T.", and "Wonder Woman" (both 2009 and 2017). Using the previous version, all of these movies failed or resulted in retrieving the incorrect movie. Everything worked correctly in the new version.

### Limitations:
None that I am aware of. If a movie did not have an IMDb id it would no longer work, though I'm not sure if that is possible.

### DIFF
--- a/couchpotato/core/media/_base/providers/torrent/yts.py
+++ b/couchpotato/core/media/_base/providers/torrent/yts.py
@@ -1,7 +1,6 @@
 from datetime import datetime
-from couchpotato.core.helpers.variable import tryInt
+from couchpotato.core.helpers.variable import tryInt, getIdentifier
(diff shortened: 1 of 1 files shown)


RuudBurger:
Thanks!

PR #7209, opened 2017-05-26
Adding file for centos service

aditaa:
Feature add

adds init file for centos systems

### Related issues:
have tested and i am running this live

### DIFF
--- a/init/couchpotato.centos.service
+++ b/init/couchpotato.centos.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=CouchPotato application instance
+After=network.target
+
+[Service]
+ExecStart=/var/lib/CouchPotatoServer/CouchPotato.py --daemon
+Type=simple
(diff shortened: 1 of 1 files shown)


RuudBurger:
How is this different then the init/couchpotato.service?

aditaa:
it mainly adds --daemon to the command 

as is it dose not work well on centos

tkimball83:
The included couchpotato.service file works fine for me on CentOS.

However, I noticed when CouchPotato automatically updates sometimes CouchPotato.py loses its executable bit.  Maybe my config is broken somewhere?

I added the following systemd override in my environment, but maybe this should be added to couchpotato.service?

ExecStartPre=/usr/bin/chmod +x /var/lib/CouchPotatoServer/CouchPotato.py

Let me know your thoughts and I'll create a PR if so.

PR #7213, opened 2017-06-03
lossless compression of images saved 4KB

Fish2:
Save Bandwidth and speed up website load time but lossless completion of all images.

doing this saved 4KB docent sound like much but why not


RuudBurger:
I already moved all images through an optimizer, not gonna add this bulk change for the 4KB, sorry.
Thanks for the PR though!

PR #7216, opened 2017-06-10
Remove automations that do not work anymore

DieBauer:
Since end of 2016 the RSS feed of moviemeter.nl is disabled (see https://www.moviemeter.nl/forum/1/8593/30#4523045).
When having the automation enabled, this causes errors in the logfile and the feature doesn't work of course.

```
6-10 13:31:30 ERROR
[edia._base.providers.base] Invalid XML returned, check "http://www.moviemeter.nl/rss/cinema" manually for issues
```

kinepolis.be website doesn't have a top 10 list anymore. (http://kinepolis.be/nl/top10-box-office/feed)

movies.io is not online anymore. (http://movies.io)


RuudBurger:
Thanks!

PR #7217, opened 2017-06-10
getInfo function needed update

jmseaton:
```
File "/apps/couchpotato2/couchpotato/core/media/movie/providers/info/omdbapi.py", line 62, in getInfo
    cached = self.getCache(cache_key, self.urls['info'] % identifier, timeout = 3, headers = {'User-Agent': Env.getIdentifier()})
TypeError: not enough arguments for format string

```
getInfo function needed update to support the api key.  Created a url variable to mirror search function.

Also fixed a typo in the info url (apikiey > apikey)


jmseaton:
sorry, i did this PR to master instead of develop, and when I tried to switch it, there were a ton of commits that showed up.

RuudBurger:
merged, thanks!

PR #7220, opened 2017-06-14
torrentday has depcreated the classic URL

jshapiro26:
the "classic" url for downloading torrent files from torrentday has been deprecated and has begun throwing 404's. As is, snatching is broken for torrentday. Changing the base_url's resolves this issue and allows snatching again.


jmonsma:
@RuudBurger this works

PR #7226, opened 2017-07-16
Update thepiratebay.py

jasonlyle88:
Remove the quotes around the search string to allow for results to be returned

Removes quotes so that thepiratebay can return results. I have had several movies not be returned because of the quoting

### Related issues:
...


RuudBurger:
Do you have an example that fails? It is there to prevent an "OR" search instead of the full title

jasonlyle88:
Couchpotato searched for: "The Silencers 1966"

Pirate Bay has: "The Silencers (Action 1966) Dean Martin 720p"

I see what you are saying. Perhaps the correct change would be to search for "{Title}" "{Year}" instead to allow for the year not always being simply listed immediately after the title. Or maybe something else of the sort. Any thoughts?

PR #7222, opened 2017-07-01
Increase Default Video Sizes - HDTS, HDCAM Never Downloads

d8ahazard:
I have been trying to figure out why HDTS and HDCAM files wouldn't
download, despite everything appearing to work properly.  This is the
problem.  Scene filesizes for TS/CAM may match the defined sizes, but
HDTS and HDCAM files are naturally larger.

If it's not ideal to increase default sizes for defined names, then
maybe HDTS and HDCAM should have their own definitions - although this
may conflict with TS/CAM?

### Description of what this fixes:
HDTS and HDCAM files were never found by the searcher.

### Related issues:
...

PR #7230, opened 2017-07-24
Add HDCam, HDTS, HDSCR profiles

d8ahazard:
Add quality profiles for HD rips of prereleases.  Currently, the
searcher will fail if the file size for releases commonly designated
with a HD- tag is larger than 1g/1.6g, so this adds support OOB for
those size releases.

If a CAM or TS is larger than 1G, it will never be downloaded by Couchpotato.  On investigation, this is almost always happening with files labeled as "HDTS" or "HDCAM".  While these aren't always the most desirable releases, the app does still provide support for them, so it'd be nice to not have to modify profiles in order to be able to have CP work as expected.

### Related issues:

Re-working of issue #7222


HazCod:
A bit unrelated but since the forum is down: 2160p isn't appearing in Qualities.

PR #7234, opened 2017-07-30
Switch from qstatus (deprecated) to queue API call for SABnzbd

Safihre:
@RuudBurger Will people that have the installer from https://couchpota.to/ also get this update?
Because we wanted to get rid of `qstatus` for a long time, but didn't because of CouchPotato.

(now correctly against `develop` and not `master`)

### DIFF
--- a/couchpotato/core/downloaders/sabnzbd.py
+++ b/couchpotato/core/downloaders/sabnzbd.py
@@ -100,7 +100,7 @@ class Sabnzbd(DownloaderBase):
 
             # the version check will work even with wrong api key, so we need the next check as well
             sab_data = self.call({
-                'mode': 'qstatus',
+                'mode': 'queue',
             })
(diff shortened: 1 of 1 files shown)


clinton-hall:
@Safihre 

The build versions (OSX and Windows) will eventually get all merged pull requests, but the  last build was 1 Sep 2015. So I'm not sure when @RuudBurger will create the next release.

On couchpota.to forums I am advising users on OSX and Windows to install from source.

Safihre:
Can't someone else do this releasing? Still all relying on 1 person? 
Why?

Safihre:
What about those (on average) 600 users that download this very much outdated version of CouchPotato every day? 
http://www.safihre.com/sabnzbd/stats.html?repo=CouchPotato/CouchPotatoServer
They're all supposed to switch to source?

PR #7236, opened 2017-08-19
Fix login problems with BitHDTV provider

fiLLLip:
BitHDTV scraping and downloading was broken after implementation of captcha. Uses cookies set by user instead of username and password. Cookies must be fetched from a active session.

Must use cookie values from logged in session:
image

### Related issues:
...


fiLLLip:
Is this repo still alive, @RuudBurger?

PR #7237, opened 2017-08-23
Replace Torrentz with Torrentz2

yldepm:
torrentz.eu has been down for a whilr. Replacing it with torrentz2 with the following limitations:
- Query URLs are different
- Cannot get "verified" quality from feed/search with torrentz2
- Fix cases when size is in GB
- cannot pass size in query (see https://torrentz2.eu/help and I have verified)

It has been tested with few scenarios


yldepm:
merge fail, will try again

PR #7238, opened 2017-08-23
Replace Torrentz with Torrentz2

yldepm:
torrentz.eu has been down for a while. Replacing it with torrentz2 with the following limitations:

- Query URLs are different
- Cannot get "verified" quality from feed/search with torrentz2
- Fix cases when size is in GB or KB
- cannot pass size in query (see https://torrentz2.eu/help and I have verified)

It has been tested with few scenarios

### DIFF
--- a/couchpotato/core/media/_base/providers/torrent/torrentz.py
+++ b/couchpotato/core/media/_base/providers/torrent/torrentz.py
@@ -15,24 +15,18 @@ log = CPLog(__name__)
 class Base(TorrentMagnetProvider, RSS):
 
     urls = {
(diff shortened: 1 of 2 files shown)


yldepm:
Hum, is anyone maintaining a fork of couchpotato?

benjaminfauchald:
Hmmm... maybe someone should. Seems like bug reporting link is down as well

jonlguk:
Hi you can just go to [Search Torrentz2.eu](http://searchtorrentz2.eu) to search all torrents

PR #7240, opened 2017-09-13
Reworked the categories for TorrentLeech to catch WEB-DL, WEBRip, HDRip and 2160p

whitter:
Noticed that CouchPotato was snatching few movies than it used to.  Seems this is down to TorrentLeech and their search categories.

CouchPotato was never seeing HDRips or WEB-DL for the brrip quality as they are located in TorrentLeech categories that where not being searched.  Noticed that WEBRip where not being looked for nor where 2160p qualities.

Had this change running for the past week and now I'm drowning in movies.


couchpotato/core/media/movie/providers/torrent/torrentleech.py
```diff
@@ -11,12 +11,14 @@
 class TorrentLeech(MovieProvider, Base):
 
     cat_ids = [
-        ([13], ['720p', '1080p', 'bd50']),
+        ([41], ['2160p']),
```
nenladar: Please add category 47 here as well.

PR #7251, opened 2017-10-22
Don't rename subtitles when there's only one language

thijstijsma:

I looked into the issue I had with Plex not detecting subtitles.

Soon I found out CouchPotato is renaming a pair of idx/sub to MovieName.idx and MovieName.en.sub, even though they are the only files.


thijstijsma:
What should this functionality do?

I've made this assumption, but there is something missing:

- If the subtitle file contains one language, it shouldn't rename.
- If the subtitle file contains multiple language, it shouldn't rename.
- If there are multiple subtitle files, it should rename each of them? But what if they also have multiple languages?

I'll adjust the code accordingly.

PR #7257, opened 2018-01-04
Update scanner.py

PeroqDev:
Add support of .mov extension in the renamer
Based on this old post in the forum: https://couchpota.to/forum/viewtopic.php?f=5&t=4642&p=22300#p22313
Tested with single .mov file that has been correctly processed by the renamer


RuudBurger:
I rather keep the .mov extension for trailers, which it is more often.
Thanks for the PR though!

PR #7259, opened 2018-01-10
Change search term for 2160p material from PTP provider

arebokert:
PTP does not return any results when searching for the term 4K. PTP uses 2160p as search term for its 4K material and so, the search term needs to be changed.


gaganpreet:
I had the same issue and can confirm this fix works. Thanks!

RuudBurger:
Thanks!

PR #7262, opened 2018-01-24
Add reverse symlink support to renamer

michalskrivanek:
adding support for reverse symlinks - move the original file to destination and create a new symlink in the original location leading to destination.
Useful alternative to hardlinks when you have different source and target filesystems and can't use them


RuudBurger:
Thanks!

PR #7268, opened 2018-02-23
updatet hepiratebay.py .. fix problematic proxies

apo-mak:
i remove proxy's that give malware infected torrent or nonexistent ones .
       ( 'httpx://pirateproxy.cat',
        'httpx://theproxypirate.pw',
        'httpx://arrr.xyz',)

the first two redirect to pirateproxy.sh that show fake malware infected torrent.
example: 
https://pirateproxy.sh/torrent/20242390/Ant-Man_and_the_Wasp_2018_HDRip_XviD_AC3-EVO

the arrr.xyz give a Error 1016 Origin DNS error .

...

### Related issues:
...

### DIFF
--- a/couchpotato/core/media/_base/providers/torrent/thepiratebay.py
+++ b/couchpotato/core/media/_base/providers/torrent/thepiratebay.py
@@ -25,7 +25,6 @@ class Base(TorrentMagnetProvider):
     http_time_between_calls = 0
 
     proxy_list = [
-        'https://pirateproxy.cat',
         'https://pirateproxy.wf',
         'https://pirateproxy.tf',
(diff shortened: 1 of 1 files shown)


nydave69:
Didn't know where else to put this...
The results returned from the pirate bay seems to match just about everything.  I have to keep ignoring results..
The bigger problem is that it downloads them and they don't match on rename so it does it again.

image

It's like this for every wanted item.  It will also match software and TV shows..
Some things it matches just doesn't make sense as to why it matched.
The image shows one of many listed, I ignore them if I can catch them before it downloaded. It will also try to download the same file over and over.  This issue is only with the pirate bay,

PR #7269, opened 2018-03-01
 Fix CP Not Auto Snatching Bug

alonalbert:
Workaround for a bug in https://api.couchpota.to/eta/<imdbid>

CP uses this API to retrieve movie release dates. The API always returns

"{dvd":0,"theater":0,"bluray":false,"expires":<some-bogus-time-about-11-days-in-the future>}

Because of this, CP will never auto-snatch movies because it can't establish a valid ETA.

This workaround tries to extract a valid release-date from the 'released' field of the media info which seems to be valid but seems to vary in format.

If it fails to convert the date, it uses the expires value.

### DIFF
--- a/couchpotato/core/media/movie/_base/main.py
+++ b/couchpotato/core/media/movie/_base/main.py
@@ -1,5 +1,6 @@
 import traceback
 import time
+from datetime import datetime
 
 from CodernityDB.database import RecordNotFound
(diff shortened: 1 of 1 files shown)


JoshDi:
this should be merged as it fixes a bug with the current master branch

PR #7296, opened 2019-04-10
Fix bug in IMDB watchlist loader

iasmanis:
Watchlist loader gets in endless loop loading list of more than 320 items

### Related issues:
None

### DIFF
--- a/couchpotato/core/media/movie/providers/automation/imdb.py
+++ b/couchpotato/core/media/movie/providers/automation/imdb.py
@@ -45,26 +45,28 @@ class IMDBBase(Automation, RSS):
     }
 
     def getInfo(self, imdb_id):
-        return fireEvent('movie.info', identifier = imdb_id, extended = False, adding = False, merge = True)
+        return fireEvent('movie.info', identifier=imdb_id, extended=False, adding=False, merge=True)
 
     def getFromURL(self, url):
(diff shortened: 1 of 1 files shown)


dacabdi:
@iasmanis IMDB's interface changed slightly, as of now the list is lazy loaded at a pace of a 100 items at a time. The scrapper is incapable to passing pages. I have a list with about ~3k films. I found a workaround by changing the HTTP query to use the key `page` and iterate through. However, watchlists, located by user id, do not allow this, you would need the actual list's URL. Could not find any other solution. Let me know if you want to incorporate my source to this.

shtrom:
I just tried this PR on top of the current develop, and it did fix my issue straight away. CouchPotato went from using 100% CPU to almost nothing (200+ titles in watchlist), and I suspect it also stopped queueing duplicates of the same titles.

@RuudBurger I think this should be merged soon.

shtrom:
(Though, maybe all the whitespace changes should be separated or removed, first).

shtrom:
@RuudBurger is there something I could do to help push this PR along?

It was working fine for me before, but I realised that I was no longer on `master` and therefore not getting upgrades. I just pulled latest `master`, and noticed that the bug was still present, then cherry-picked commit 1303955874cae38e601f3b4a88a580973e538787 on top, and this fixed the high CPU load as before.

PR #7297, opened 2019-06-01
Fix 4K qualities + rarbg 4k support

gacopl:
This one fixes
* qualities scoring, before a lot movies could end up being marked as 2160p while they actually were 1080p, and some 2160p web-dl rips were not recognized as 2160p but brrip or totally ommitted
* fix rarbg as it has new categories for 4k and 4k hdr and releases present there are not always present in 'movies' category aswell

After those fixes my CP is accurate like never before

### Related issues:
...


Namlaz:
I would suggest to leave h264/x264 in 2160p quality in addition to adding h265/x265

gacopl:
when x264 left i got many false positives with actual 1080p blurays

nydave69:
I didn't know where else to put this.....

CP is matching just about everything now including TV shows and software for Movies....
It will download them and when of course they don't match what should have been downloaded it does it again and again.

image

PR #7310, opened 2019-12-27
Added YTS Popular Automation.

rhyswilliamsza:
**_NEW FEATURE_**
YTS Popular-Downloads grabbing: Added YTS automation provider, pulling popular-downloads directly from their site. Will safely and automatically fail if website unavailable or layout changes.

### DIFF
--- a/couchpotato/core/media/movie/providers/automation/yifypopular.py
+++ b/couchpotato/core/media/movie/providers/automation/yifypopular.py
@@ -1,5 +1,4 @@
-import requests
-from html.parser import HTMLParser
+import HTMLParser
 from couchpotato import fireEvent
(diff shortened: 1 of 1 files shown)


rhyswilliamsza:
Requires HTML parser module. This is included by default with Python 2.7+. Unsure why travis couldn't find the library.

nydave69:
2 questions, 
1.  how does one add this to the current master that I'm running..
2. Could this be done for RARBG..

rhyswilliamsza:
I can quickly write a RARBG version if you'd like @nydave69?

nydave69:
That would be great, since i see them releasing a lot of new things.
merge it with master...

PR #7318, opened 2020-02-03
Use 7 letter IMDB ID's unless title has an 8 char long IMDB ID.

PaulW:
IMDB still returns 7-char ID's on older titles when searching, simply
blanketin to 8 chars all the time (even if IMDB support 8-char padded
ID's) will ultimately fail lookups on other services.

This allows us to default to a 7-char ID, padded with zero's to ensure
that it still works as expected, but ultimately allow newer 8-char ID's
as-and-when required for newer titles.

Another side effect observed, is that if changing from 7 to 8char as previously done, collections can suffer from duplications, and the Wanted List will repopulate with films which have already downloaded, but with an 8-digit IMDB ID.

### DIFF
--- a/couchpotato/core/helpers/variable.py
+++ b/couchpotato/core/helpers/variable.py
@@ -197,9 +197,9 @@ def getImdb(txt, check_inside = False, multiple = False):
         ids = re.findall('(tt\d{4,8})', txt)
 
         if multiple:
-            return removeDuplicate(['tt%08d' % tryInt(x[2:]) for x in ids]) if len(ids) > 0 else []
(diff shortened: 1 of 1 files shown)


nydave69:
Could this be why my Manage view shows 23046 movies,
Thousands more then it was before the last upgrade...

PaulW:
Quite possibly, mine duplicated over 4000 movies itself, most of which ended up in the Wanted list even though they were already downloaded.

I also noticed a greater number of 404's when hitting api.themoviedb.org, www.omdbapi.com and also api.couchpota.to with the longer imdb_id's if they were 8 digits long against older items.  This also reduced considerably since changing the way in which they are formatted.

I've also seen no duplications of films since either.  Testing with the old forced 8char ID did cause duplications to come through.

RuudBurger:
Thanks!

PR #7319, opened 2020-02-04
Develop

RuudBurger:
...

### Related issues:
...


nydave69:
I have now thousands of movies in the managed list that I know is wrong and over 1100 in the wanted list that used to be just under 100..  I've done a full scan of the media and it has increased. the wanted list hasn't changed. The other thing is that CP has now not found one movies to download in over 2 weeks.  I figured with the changes it would have loaded movies for download even ones I have considering the counts. I can see that it is searching but doesn't seem to find matches.. Even though I know there are.

PR #7321, opened 2020-02-15
Fix YTS 7-Digit IMDB Searches

rhyswilliamsza:
YTS does not find older 7-digit films after the change in the previous update. This fixes that issue.

This will only replace tt0 with tt, thus not affecting the new 8-digit IMDB ids.


rhyswilliamsza:
Please ignore this. YTS has updated their API and this is no longer necessary.

PR #7323, opened 2020-02-21
Add YTS Popular Downloads as a Chart Option

horganhoozen:
Using code from the bluray.py file, I have added the option to display items from the YTS.MX popular downloads rss feed as a chart on the home screen.

LIMITATION - It only seems to pull 4 items at this time.


rhyswilliamsza:
Please ignore this. YTS has updated their API and this is no longer necessary.

PR #7328, opened 2021-01-16
Updated protocol from HTTP to HTTPS since IMDB returns 301 on HTTP

BartoszKaszewczuk:
Updated protocol from HTTP to HTTPS since IMDB returns 301 Moved Permanently on HTTP which causes CP to find no new movies.


BartoszKaszewczuk:
Closed because was opened against Master branch.

